PR #10, opened 2021-04-26
Automatically add permissions to Lambda functions

mnapoli:
This is WIP until we figure out how to reference resources properly.

### DIFF
--- a/src/components/Queues.ts
+++ b/src/components/Queues.ts
@@ -1,4 +1,3 @@
-import { PolicyStatement } from "@aws-cdk/aws-iam";
 import { CfnOutput, Duration } from "@aws-cdk/core";
 import chalk from "chalk";
 import { Queue } from "@aws-cdk/aws-sqs";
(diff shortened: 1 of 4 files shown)


src/components/Queues.ts
```diff
@@ -119,7 +119,15 @@ export class Queues extends Component<
...
-    async permissions(): Promise<PolicyStatement[]> {
-        return Promise.resolve([]);
+    permissions(): PolicyStatement[] {
+        const configuration = this.getConfiguration() ?? {};
+
+        return Object.entries(configuration).map(([name]) => {
+            const cfId = formatCloudFormationId(`${name}`);
+
+            return new PolicyStatement("sqs:SendMessage", [
+                cfGetAtt(`${cfId}Queue`, "Arn"),
```
mnapoli, line 129: This reference doesn't work: it will reference `XxxQueue`, whereas the CDK will generate a resource called `XxxQueueE3DFB796` (unique ID).

fredericbarthelet: Pourquoi tu utilises `cfGetAtt` ?

PR #14, opened 2021-05-08
Update documentation (readme)

seblegall:
Hi guys,

Just a few quick fix on the readme here. I could also add examples for the queues and webhook components. But I feel like the documentation should be auto-generated. Having a basic example in the readme is great. It helps people understand how it works and, in my point of view, it also helps to consider the use of the lib/plugin for their use case.

I am not very familiar with the JS/TS ecosystem. But since you're using JSON Schema to validate the components configuration, it should be easy to convert that into human-readable documentation such as Markdown. You might need to extract the JSON schema in their own files.

Anyway, for now, I chose to stick with the serverless strategy and put required and optional parameters in the example.

Cheers,

PS: an example directory with concrete implementation could be very useful too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ npm i -D serverless-lift
 ## Quick start
 
 *serverless.yml*
+
 ```yaml
 service: my-app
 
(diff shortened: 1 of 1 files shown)


mnapoli:
Hi, thanks a lot for the contribution!

So far we've been focusing on getting feedback on the design of the first components. Now that we have, we'll work towards a first beta, with full documentation (I'll be able to start on that next week). That's why so far there isn't much documentation (we were mostly focused on github discussions and articles on medium).

I think we should keep the examples on the readme very simple, just to illustrate what lift can do. And then have full fledged documentation for each component (we already have documentation drafts outside of github).

mnapoli:
I've added the documentation in #15. Sorry for closing this 😄

PR #16, opened 2021-05-11
Improve static website uploads

mnapoli:
Static websites used to be deployed via a `aws s3 sync` sub-command. That was good for prototyping, but this isn't viable for production because:

- we cannot assume the CLI is installed
- the credentials/profiles to use will not be forwarded to the sub-command

In this PR I've implemented an upload in pure JavaScript. There is room for improvements/optimizations, but I prefer shipping something that works ASAP.

It turns out the JS SDK doesn't have any `sync` mechanism. All JS packages that used to offer some kind of feature like this are all dead and all had naive implementations, like this one.

TODO:

- [x] automated tests


src/components/StaticWebsites.ts
```diff
@@ -280,11 +280,10 @@ class StaticWebsiteConstruct extends ComponentConstruct {
         }
 
         log(`Uploading directory '${this.configuration.path}' to bucket '${bucketName}'`);
-        // TODO proper upload, without going through a subcommand
-        spawnSync("aws", ["s3", "sync", "--delete", this.configuration.path, `s3://${bucketName}`], {
-            stdio: "inherit",
-        });
-        await this.clearCDNCache();
+        const changes = await s3Sync(this.serverless.getProvider("aws"), this.configuration.path, bucketName);
```
fredericbarthelet, line 283: en règle générale en ts, on préfère passer un objet d'arg et un objet de retour pour bénéficier des tooltips du typage. ton arg de retour, si c'est un boolean, je partirai sur un `hasChanges` plutôt que juste `changes`

```suggestion
        const { hasChanges } = await s3Sync({
          provider: this.serverless.getProvider("aws"),
          localPath: this.configuration.path,
          bucketName
        });
```

src/utils/s3-sync.ts
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * @return True if some changes were uploaded.
+ */
+export async function s3Sync(aws: Provider, localPath: string, bucketName: string): Promise<boolean> {
+    let hasChanges = false;
+    const filesToUpload: string[] = await listFilesRecursively(localPath);
+    const existingS3Objects = await s3ListAll(aws, bucketName);
+
+    // Upload files 5 by 5
+    for (const batch of chunk(filesToUpload, 2)) {
```
fredericbarthelet, line 35: chunk size of 2 but comment says "5 by 5" ?

src/utils/s3-sync.ts
```diff
@@ -0,0 +1,145 @@
...
+                const subFiles = await listFilesRecursively(fullPath);
+
+                return subFiles.map((subFileName) => path.join(fileName, subFileName));
+            }
+
+            return [];
+        })
+    );
+
+    return files.flat(1);
```
fredericbarthelet, line 87: use Set rather than array, you'll be much more efficient to test for the existence of an item in existing files and not in S3 and the other way around

src/utils/s3-sync.ts
```diff
@@ -0,0 +1,145 @@
...
+                return {
+                    Key: key,
+                };
+            }),
+        },
+    });
+}
+
+export function computeS3ETag(fileContent: Buffer): string {
+    return '"' + crypto.createHash("md5").update(fileContent).digest("hex") + '"';
```
fredericbarthelet, line 144:
```suggestion
    return `"${crypto.createHash("md5").update(fileContent).digest("hex")}"`;
```

package.json
```diff
@@ -51,7 +55,7 @@
     "nodemon": "^2.0.7",
     "prettier": "^2.2.1",
     "serverless": "^2.35.0",
-    "sinon": "^9.2.2",
+    "sinon": "^9.2.4",
```
fredericbarthelet, line 58: we're using jest as an assertion library, why introduce sinon ?

test/unit/staticWebsites.test.ts
```diff
@@ -219,4 +231,79 @@ describe("static websites", () => {
             },
         });
     });
+
+    it("should synchronize files to S3", async () => {
+        sinon.stub(CloudFormationHelpers, "getStackOutput").returns(Promise.resolve("bucket-name"));
```
fredericbarthelet, line 236: You should use `jest.mock` here instead -> https://jestjs.io/docs/mock-functions

mnapoli: Couldn't make it work with Jest, e.g.

image

package.json
mnapoli: It was already there, it's just the version number that is different.

src/utils/s3-sync.ts
mnapoli: I gave it a try but I'm not sure I got what you meant: https://github.com/getlift/lift/compare/s3-sync...s3-sync-set

With a Set we don't gain much, and the code ends up more complex. Is that what you had in mind?

fredericbarthelet: You can use `Set.add` api to progressively build your set in the previous `items.map` using `for item of items` instead. You can use `Set.values()` rather than  destructuring where you need to iterate over the set values (or you can `for myValue of mySet`). But if you don't see too much benefit from using it, keep it that way :)

test/unit/staticWebsites.test.ts
fredericbarthelet: manual `jest.mock` is to be done on the same scope level as the import (meaning outside of the describe block). If you want, you can merge it like this and I'll update one of the test with the corresponding jest mocking system to give you an exemple :)

mnapoli: OK I see, but the problem above (in red) is the type error. But maybe I'm missing your point, I'll check out the example later.

src/utils/s3-sync.ts
mnapoli: I gave it a quick try and I see what you meant by "use `Set.add`": https://github.com/getlift/lift/compare/s3-sync...s3-sync-set

However I must use restructuring inside `chunk()` because chunk takes an array (even `.values()` doesn't work because it doesn't return an array).

Anyway to be honest here the change is 1:1 in terms of LOC and complexity, so I'd rather merge. But that was interesting to learn about that, thanks!

PR #21, opened 2021-06-04
Add website security headers via CloudFront functions

mnapoli:
This pull request adds recommended security headers to static websites.

This feature originated from the RFC here: https://github.com/getlift/lift/discussions/5#discussioncomment-621899

It does so via the newly announced CloudFront Functions 🎉 
After playing with this, I am really happy to say that I did not see any downside to using CloudFront Functions:

- no additional time to the deployment (Functions deploy fast!)
- unlike Lambda@Edge, Functions live in the CloudFront console and don't pollute the Lambda console over many different regions
- unlike Lambda@Edge, logs are centralized in 1 region (though we don't log anything in our functions)
- Functions execute extremely fast

In other words, it just works. And since 2 days ago they are now supported by the CDK.

Please review the security headers I have added. I am not an expert, I have followed recommendations but I may be wrong.

I am also interested in the impacts it may have on users: could those security headers make the life of developers harder? Could it create issues? If so, please let me know so that I can document them (and document workarounds).

Thank you @guillaumem-theodo for contributing to the RFC. Review is very welcome here :)

TODO:

- [x] tests
- [x] documentation


package.json
```diff
@@ -5,17 +5,17 @@
...
+    "@aws-cdk/aws-apigatewayv2-integrations": "1.107",
+    "@aws-cdk/aws-cloudfront": "1.107.0",
+    "@aws-cdk/aws-cloudwatch": "1.107.0",
+    "@aws-cdk/aws-events": "1.107.0",
+    "@aws-cdk/aws-iam": "1.107.0",
+    "@aws-cdk/aws-lambda": "1.107.0",
+    "@aws-cdk/aws-s3": "1.107.0",
+    "@aws-cdk/aws-sns": "1.107.0",
+    "@aws-cdk/aws-sqs": "1.107.0",
+    "@aws-cdk/core": "1.107.0",
```
mnapoli, line 18: I had to upgrade to get the latest CDK version that supports Functions.

src/components/StaticWebsite.ts
```diff
@@ -77,7 +79,26 @@ export class StaticWebsite extends CdkConstruct implements Construct {
...
+        /**
+         * CloudFront function that manipulates the HTTP responses to add security headers.
+         */
+        const code = `function handler(event) {
+    var response = event.response;
+    response.headers = Object.assign({}, {
+        'x-frame-options': {value: 'SAMEORIGIN'},
+        'x-content-type-options': {value: 'nosniff'},
+        'x-xss-protection': {value: '1; mode=block'},
+        'strict-transport-security': {value: 'max-age=63072000'},
```
mnapoli, line 91: Here are the security headers being added to the response.

src/components/StaticWebsite.ts
```diff
@@ -77,7 +79,26 @@ export class StaticWebsite extends CdkConstruct implements Construct {
...
+    }, response.headers);
+    return response;
+}`;
+        const edgeFunction = new cloudfront.Function(this, "EdgeFunction", {
+            functionName: `${this.provider.stackName}-${this.provider.region}-${id}-response`,
+            code: cloudfront.FunctionCode.fromInline(code),
+        });
+
         const distribution = new CloudFrontWebDistribution(this, "CDN", {
+            comment: `${provider.stackName} ${id} website CDN`,
```
mnapoli, line 101: Just to help users figure things out in the console:

image

t-richard:
This is a great move and some nice defaults.

But IMHO this should be configurable but still enabled by default. Maybe enabling all/nothing (`securityHeaders: true/false`) is enough for now and could improved with finer-grained config later on if requested often.

Otherwise some users might have a hard time understanding why it is not working as they expect (like if they need to use iframes).

Also a small part in the docs to state what headers are enabled by default would be great.

src/components/StaticWebsite.ts
```diff
@@ -77,7 +79,26 @@ export class StaticWebsite extends CdkConstruct implements Construct {
             comment: `Identity that represents CloudFront for the ${id} static website.`,
         });
 
+        /**
+         * CloudFront function that manipulates the HTTP responses to add security headers.
+         */
+        const code = `function handler(event) {
```
M1ke, line 85: Could we import this content from a separate file? Means if someone is contributing they get a proper JS file to avoid things like syntax errors which are harder to spot when it's a string?

src/components/StaticWebsite.ts
M1ke: Do we want to consider extending this with `; includeSubDomains; preload`? It adds more risk (if someone is unaware, they can cause all subdomains to be required to use HTTPS) but also makes a production application more secure (as browsers can then force the domain to HTTPS internally even if a user has never visited it). Maybe a mode setting?

src/components/StaticWebsite.ts
mnapoli: Good point, but now that the code is dynamic that makes a bit less sense I think (I've updated the code).

mnapoli:
@t-richard that is a really good point. I've reviewed the headers again and _I think_ the only practical downside for users is that they cannot use the website in an iframe.

So I have added an option just for that:

```
landing:
  ...
  security:
    allowIframe: true
```

In the future, we could even allow passing a domain name (for better security), but I'll relegate that to later.

AFAICT other headers are safe to always add, so why not keep them on all the time.

Also in the future, we may want another option to allow inserting _any_ header to responses. I've built the code to accept that change easily. In this PR however, I'll keep it simple and focus on a few specific headers.

src/components/StaticWebsite.ts
mnapoli: That's a good point, but I would be cautious about adding that right now. Indeed, here we control the static website, but users could host anything in subdomains, so I wouldn't want to break that.

Adding an option is probably the best course of action. In that case, I'd focus on covering the main use case first and add the option later.

PR #23, opened 2021-06-07
[RFC] Deploy any CDK construct

mnapoli:
This is a draft to discuss supporting generic CDK constructs. Here is how it can be used so far:

```yaml
service: app

provider:
    name: aws

constructs:
    cron:
        type: cdk
        file: ./construct

plugins:
    - serverless-lift
```

And the construct would be defined here in `construct.js` for example:

```js
const events = require('@aws-cdk/aws-events');
const targets = require('@aws-cdk/aws-events-targets');
const lambda = require('@aws-cdk/aws-lambda');
const cdk = require('@aws-cdk/core');

const fs = require('fs');

class Cron extends cdk.Construct {
    constructor(app, id) {
        super(app, id);

        const lambdaFn = new lambda.Function(this, 'Handler', {
            code: new lambda.InlineCode(fs.readFileSync('lambda-handler.py', { encoding: 'utf-8' })),
            handler: 'index.main',
            timeout: cdk.Duration.seconds(3),
            runtime: lambda.Runtime.PYTHON_3_6,
        });

        const rule = new events.Rule(this, 'Rule', {
            schedule: events.Schedule.expression('rate(5 minutes)')
        });

        rule.addTarget(new targets.LambdaFunction(lambdaFn));
    }
}

module.exports = Cron
```

(example taken from https://github.com/aws-samples/aws-cdk-examples/tree/master/typescript/lambda-cron)

This isn't meant to be merged as-is, but rather to discuss the feature.

What I've seen so far:

- would work mostly with high-level components, where we have little need for customization (props for example), because the CDK is mostly meant to be used programmatically (we can't call methods in YAML)
- so far no support for exposing references & outputs, which confirms the line above: it's a good fit for high-level constructs, not assembling multiple constructs in serverless.yml
- no automatic permissions (unlike other Lift components)
- no support for uploading assets (e.g. Lambda code) because we can't use the CDK's bootstrap stack (CDK toolkit), so some constructs wouldn't work

I'm not 100% sure this is worth it in its current state TBH.

### DIFF
--- a/src/components/CdkConstruct.ts
+++ b/src/components/CdkConstruct.ts
@@ -1,0 +1,36 @@
+import * as cdk from "@aws-cdk/core";
+import { FromSchema } from "json-schema-to-ts";
+import * as path from "path";
+import Construct from "../classes/Construct";
+
+export const CDK_CONSTRUCT_DEFINITION = {
+    type: "object",
+    properties: {
+        type: { const: "cdk" },
+        file: { type: "string" },
+    },
+    required: ["type", "file"],
+} as const;
+
+export class CdkConstruct implements Construct {
+    constructor(scope: cdk.Construct, id: string, configuration: FromSchema<typeof CDK_CONSTRUCT_DEFINITION>) {
+        const props: Record<string, any> = Object.assign({}, configuration);
+        delete props.type;
+        delete props.file;
+
+        const C = require(path.join(process.cwd(), configuration.file));
+        new C(scope, id, props);
+    }
+
+    references(): Record<string, Record<string, unknown>> {
+        return {};
+    }
+
+    commands(): Record<string, () => void | Promise<void>> {
+        return {};
+    }
+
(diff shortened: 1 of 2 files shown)


t-richard:
I like the idea of it being extensible but this looks premature to me.

As you pointed out, there are numerous limitations that are making it only a good fit to write only very specific extensions.

Also, this is a beta version and you recently broke the config to go from components to constructs. This is likely to happen again as you evolve the project into something bigger and better. Breaking this kind of thing would be frustrating for most people. (just to be clear, I'm not blaming you for breaking things here :sweat_smile:)

Last, but that is my personal POV : if I need something more "custom", I would turn to Cloudformation or plain serverless Framework if possible (like in your CRON example). Or contribute it :blush: 
But that may be related to me not being confortable enough with the CDK.

mnapoli:
Replaced by #52

PR #24, opened 2021-06-07
Add a command and documentation to eject from Lift

mnapoli:
This PR adds a new `serverless lift eject` command. It also adds the related documentation.

This allows users to eject from Lift (to CloudFormation) easily.

Preview of the command:

Screen 20210607 KWNKp2nc@2x
Screen 20210607 RW7cuLnD@2x

### DIFF
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -40,7 +40,7 @@ class LiftPlugin {
 
         this.serverless = serverless;
 
-        this.commands["lift"] = {
+        this.commands.lift = {
             commands: {
                 eject: {
                     lifecycleEvents: ["eject"],
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -5,6 +5,7 @@ import { JSONSchema6 } from "json-schema";
 import { AwsIamPolicyStatements } from "@serverless/typescript";
 import * as path from "path";
 import { readFileSync } from "fs";
+import { dump } from "js-yaml";
 import type {
     CloudformationTemplate,
     CommandsDefinition,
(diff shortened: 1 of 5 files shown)


t-richard:
Love the idea of promoting the ejectability nature of Lift, even if i'd rather not have to do it.

Maybe that's next step because it would imply a lot more work I suppose, but it would be awesome to split cloudformation for each construct like this
```
"upload" storage construct :
<only relevant cloudformation>

"heavy-lifting" queue construct :
<only relevant cloudformation>
....
```

This way would make it possible to eject single constructs easily.

But that would cause other problems with webhooks for example because it's a shared EventBridge for all webhooks IIRC so it could confuse things even more.

Also you are outputting the template in JSON but by default Serverless Framework works with YAML so users would have to transform the JSON to YAML. Is there a way to output YAML directly ?

But maybe users wanting to eject are supposedly saavy enought to know how to do all of that by themselves and the current implementation is good enough :smile: 

LMKWYT

mnapoli:
> it would be awesome to split cloudformation for each construct

Yes! This is what I initially tried to implement (it shouldn't be too hard I believe because CloudFormation IDs are prefixed). But that falls short of ejecting everything because some constructs add extra resources elsewhere, for example IAM permissions, Lambda functions, shared resources (as you mention)…

So this is why I figured "let's eject from Lift entirely". Then users can figure out what they prefer to do, and this is a bit more effort on them but it's not likely to happen every day.

> Also you are outputting the template in JSON but by default Serverless Framework works with YAML so users would have to transform the JSON to YAML. Is there a way to output YAML directly ?

Oh good point! I'll fix that in the PR, that shouldn't be too much effort and it will be much better for users.

> maybe that could be sorted out more easily with tags ?

Might not be worth it for this feature, but this is a good idea! I'm not sure if it could have other uses, but it's a good one to keep in mind.

src/plugin.ts
```diff
@@ -36,12 +40,21 @@ class LiftPlugin {
 
         this.serverless = serverless;
 
+        this.commands["lift"] = {
```
fredericbarthelet, line 43:
```suggestion
        this.commands.lift = {
```

fredericbarthelet (approved):

t-richard (approved):

PR #25, opened 2021-06-08
[WIP] Improve readme lisibility

fredericbarthelet:
Understand in a single look what Lift does and how to use it.
Start with Lift main value and then dive in detailed set of features offered

PR #26, opened 2021-06-08
Simplify README text content

mnapoli:
On top of #25, this PR reworks the "text" part of the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,74 +1,103 @@
 ![](docs/lift.png)
 
-*Shipping Serverless features made easy*
+Lift is a plugin that leverages the AWS CDK to expand the [Serverless Framework](https://www.serverless.com/) beyond functions.
 
-## Features
+Deploy production-ready websites, queues, storage buckets and more with a few lines in serverless.yml.
 
-Lift is a [Serverless](https://www.serverless.com/) plugin that simplifies deploying well-architected serverless applications.
-
-It is made for developers new to serverless, as well as experienced AWS builders.
-
-- ⚡️ **Get started with Serverless** - Developer-friendly vocabulary, no AWS knowledge required
-- 📦 **Faster delivery** - Production-ready components with minimal configuration
-- 🔁 **Not invasive** - Integrates with existing Serverless Framework projects
-- ⏏️ **No lock-in** - Eject to CloudFormation at any time
+⚡️ **For developers** - No AWS knowledge required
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,74 +1,108 @@
...
 ## Installation
 
+Lift is a [Serverless Framework plugin](https://www.serverless.com/plugins/), install it in your project:
+
 ```bash
-npm i -D serverless-lift
+serverless plugin install -n serverless-lift
 ```
 
+If you prefer, you can install via `npm` directly: `npm i -D serverless-lift`.
```
t-richard, line 22: I always find the short syntax confusing, especially for non JS devs, which serverless framework users may not be.

And most of the time it would be copy-pasted anyway.

```suggestion
If you prefer, you can install via `npm` directly: `npm install --save-dev serverless-lift`.
```

t-richard (commented):
Also I would reorganize the docs folder to make it easier to browse. 

When testing Lift I wanted to be able to go in the `docs` folder and have one folder per component.

This would allow for easier browsing of the docs and maintainance when adding more components I think.

Also renaming files to README.md in each folder would make those rendered by default in Gihub so you could browse to `docs/queue` and see the formatted Markdown there.

README.md
mnapoli: Good point. I wasn't familiar with JS for a long time and now that you mention it, I remember being confused/worried with these unknown options.

mnapoli:
> Also I would reorganize the docs folder to make it easier to browse.

👍  yes I agree too.

However having 1 folder per component is probably a bit overkill and unpractical, as you would have to move up and down the tree to switch to another component.

That being said, we can still clean up that directory, maybe something like this:

```
comparison.md
constructs/
  storage.md
  queue.md
  webhook.md
  static-website.md
  img/
    images go here
```

t-richard:
> However having 1 folder per component is probably a bit overkill and unpractical, as you would have to move up and down the tree to switch to another component.

When browsing Github I use the breadcrumb A LOT so I would just click on `docs` in there, but that's maybe just me :sweat_smile:  

Moving images to a subfolder would be a good move too and would make the docs folder less cluttered. 
Most users will use the main README to navigate anyway so let's not waste time on this for now :slightly_smiling_face:

mnapoli:
Will do adjustments on master to prepare for v1 release.

PR #31, opened 2021-06-14
Instantiate constructs after all variables are resolved

### LINKED ISSUE #29: `${param}` variables are not resolved in constructs
I just spent several hours trying to understand why the parameters from the serverless dashboard weren't correctly being passed to lift plugin until I came across this note on the [serverless plugins' guide](https://www.serverless.com/framework/docs/providers/aws/guide/plugins/).

> Note: Variable references in the serverless instance are not resolved before a Plugin's constructor is called, so if you need these, make sure to wait to access those from your hooks.

Which means that trying to do the following will result in an error "_ARNs must start with "arn:" and have at least 6 components: ${param:CERTIFICATE}_".

```yaml
constructs:
  app:
    domain: ${param:DOMAIN}
    certificate: ${param:CERTIFICATE}
    type: static-website
    path: dist
```

I fixed it on a fork moving the Constructs loader to the initialize hook https://github.com/pedrodim/lift/commit/1c476382d38236bd9ae96eacbe6cbb40859ab603, which seems to work fine for the Static website construct (it does correctly deploy my website without errors) but obviously now some tests are broken and I have no idea how to fix those.

mnapoli:
This should fix #29.

On master:

- constructs are instantiated in the constructor of the Lift plugin
- that means that variables resolved _after_ the Lift plugin is instantiated are not resolved in constructs

In this branch:

- constructs are instantiated in the `initialize` hook (at this point, all variables are resolved)
- to achieve that, commands are now defined statically
- and Lift variables are resolved lazily using CDK tokens (CDK lazy variables)

TODO:

- [ ] additional testing


mnapoli:
OK now I'm facing the problem described in #8 again (the failing test). Will continue to think about this.

mnapoli:
Idea: try implementing an `asyncInit()` method (https://github.com/serverless/serverless/blob/c563581ac98764edf653c1a5337d1b7d2b61ea63/lib/classes/PluginManager.js#L751-L753) instead of creating constructs in `initialize()`.

mnapoli:
Replaced by #50

PR #35, opened 2021-06-16
Implements the deprecated `variableResolvers` to make sure we display useful error messages to users

mnapoli:
See for background https://github.com/serverless/serverless/issues/9610

Now when a variable fails to be resolved, the users see the actual error.

I think that could even be improved by using the `ServerlessError` class, but I have that in a separate branch for now.

Will add tests of variables separately from that.

### DIFF
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -9,6 +9,7 @@ import { FromSchema } from "json-schema-to-ts";
 import type {
     CloudformationTemplate,
     CommandsDefinition,
+    DeprecatedVariableResolver,
     Hook,
     Serverless,
     VariableResolver,
@@ -56,8 +57,8 @@ class LiftPlugin {
     public readonly stack: Stack;
     public readonly hooks: Record<string, Hook>;
     public readonly commands: CommandsDefinition = {};
-    public readonly configurationVariablesSources: Record<string, VariableResolver> = {};
-    public variableResolvers: any;
+    public readonly configurationVariablesSources: Record<string, VariableResolver>;
+    public readonly variableResolvers: Record<string, DeprecatedVariableResolver>;
 
     constructor(serverless: Serverless) {
         this.app = new App();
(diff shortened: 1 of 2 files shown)


src/plugin.ts
```diff
@@ -57,6 +57,7 @@ class LiftPlugin {
     public readonly hooks: Record<string, Hook>;
     public readonly commands: CommandsDefinition = {};
     public readonly configurationVariablesSources: Record<string, VariableResolver> = {};
+    public variableResolvers: any;
```
mnapoli, line 60: TODO

src/plugin.ts
```diff
@@ -82,14 +84,18 @@ class LiftPlugin {
...
-            // TODO these 2 variable sources should be merged eventually
             construct: {
                 resolve: this.resolveReference.bind(this),
             },
         };
+        this.variableResolvers = {
+            construct: (fullVariable) => {
+                const address = fullVariable.split(":")[1];
+
+                return Promise.resolve(this.resolveReference({ address }).value);
```
fredericbarthelet, line 96: does this need to be async ? 
```suggestion
                return this.resolveReference({ address }).value;
```

src/types/serverless.ts
```diff
@@ -19,6 +19,7 @@ export type VariableResolver = {
         options: Record<string, string>;
     }) => { value: string | Record<string, unknown> } | Promise<{ value: string | Record<string, unknown> }>;
 };
+export type DeprecatedVariableResolver = (variable: string) => Promise<Record<string, unknown>>;
```
fredericbarthelet, line 22: does it need to be async ?
```suggestion
export type DeprecatedVariableResolver = (variable: string) => Record<string, unknown>;
```

src/plugin.ts
mnapoli: Yes it does, at least that's how other variable resolvers seem to be implemented, and things didn't work when I did not return a promise.

src/types/serverless.ts
mnapoli: Yep

Issue #38, opened 2021-06-17
Node 10 compatibility ("flat is not a function")


Aikoze:
Hello :hand: 

I have probleme using  `<construct>:upload` , I'm trying to deploy my landing page using the configuration said on documentation

I run :  
`serverless landing:upload`

With this config bellow :

```
constructs:
    landing:
        type: static-website
        path: landingPage

```

And this error append

![image](https://user-images.githubusercontent.com/23453566/122402838-25169d00-cf7e-11eb-83da-820af7f40593.png)

Seems like i don't have permission but I'm also using serverless to deploy my API using bref.sh with the same IAM role.

mnapoli:
Hey @Aikoze, thank you for trying out Lift, and sorry about the poor experience here!

We just discovered today that Lift doesn't work well with Node 10: `.flat()` is a function available in Node 12.

We are looking into configuring our builds to publish versions compatible with Node 12, but in the meantime if you have any way to upgrade Node that should work better.

We'll let you know!

Aikoze:
I upgraded to Node v12 and everything is fine

mnapoli:
@Aikoze thanks for the update!

PR #46, opened 2021-06-18
Fix node10 compatibility

adriencaccia:
Use lodash flatten instead of Array.prototype.flat to fix the issue.

To prevent future issues, I lowered ts lib to match runtime requirements.

Fix: #38

### DIFF
--- a/src/constructs/StaticWebsite.ts
+++ b/src/constructs/StaticWebsite.ts
@@ -21,6 +21,7 @@ import chalk from "chalk";
 import { CreateInvalidationRequest, CreateInvalidationResult } from "aws-sdk/clients/cloudfront";
 import { S3Origin } from "@aws-cdk/aws-cloudfront-origins";
 import * as acm from "@aws-cdk/aws-certificatemanager";
+import { flatten } from "lodash";
 import { log } from "../utils/logger";
 import { s3Sync } from "../utils/s3-sync";
 import AwsProvider from "../classes/AwsProvider";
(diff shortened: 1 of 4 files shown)


t-richard:
I'm not sure node 10 should be supported. 

Node 10 is EOL since a couple of months and AWS support is ending in August 2021. See [this AWS docs page](https://docs.amazonaws.cn/en_us/lambda/latest/dg/runtime-support-policy.html).

Why do you think this is needed ?

mnapoli:
Thanks a lot @adriencaccia for starting this!

I'm a bit torn too. I'll try to see internally if we have any data about Node 10 usage, to see if we should support it.

The thing is that this PR doesn't fix all the problems (AFAICT): we use `flat()` in several places, and it's not impossible there are other places too where we use code not compatible with Node 10.

adriencaccia:
> The thing is that this PR doesn't fix all the problems (AFAICT): we use `flat()` in several places,

This is my bad, I forgot to lint and type check my code before submitting the PR. I force-pushed changes to change all usage of `flat()`.

> it's not impossible there are other places too where we use code not compatible with Node 10.

Thanks to my second commit, both `eslint` and `tsc` will now throw errors when using code not compatible with node 10. That is why the CI failed when I first submitted the PR.

> Why do you think this is needed ?

This PR intent to make it possible for users to use the plugin with node 10, even if there are not that many.
As you can see, the cost to reach compatibility is not that high, here we only have to stop using `flat()` and use lodash's `flatten`.
And by lowering TS lib to `es2017`, we assure that code that would break compatibility will not be introduced.

mnapoli:
Hey @adriencaccia, the `lib: "es2017"` seems to be a very good solution indeed! Just tried it locally and it showed the errors as you mentioned. I learned something cool today 🙏 

I really like your PR: we can support Node 10 with very little effort, it seems like a no-brainer to merge. And the CI validation is perfect.

Thanks a lot!

PR #48, opened 2021-06-19
Add VS Code settings and automatic file formatting

adriencaccia:
Hello !

While developing on the projet, I noticed file formatting through `eslint` and `prettier` was not automated, and `prettier` format was not enforced in the CI.

This PR adds:

- a VS Code config with settings to automatically format and run `eslint` on saved files
- a VS Code list of recommended extensions
- a new test in the CI which ensure that all files are correctly formatted with `prettier`
- automatic formatting of staged files


.eslintrc
```diff
@@ -1,106 +1,77 @@
...
-            "external",
-            "builtin"
-          ],
-          "internal",
-          [
-            "parent",
-            "sibling",
-            "index"
-          ]
+    "extends": ["eslint:recommended", "plugin:prettier/recommended"],
```
mnapoli, line 2: Can you confirm that there are no other changes in this file than code formatting?

.github/CONTRIBUTING.md
```diff
@@ -4,13 +4,13 @@ Thank you for wanting to contribute!
 
 Before implementing anything major:
 
-- For new components, [open a GitHub discussion](https://github.com/getlift/lift/discussions/categories/components) and explain the use case with an example
-- For new features, [open a GitHub issue](https://github.com/getlift/lift/issues/new) and explain the use case with an example
+-   For new components, [open a GitHub discussion](https://github.com/getlift/lift/discussions/categories/components) and explain the use case with an example
```
mnapoli, line 7: Could you exclude documentation (Markdown files) from linting and revert these changes?

I feel like changes in these files aren't ideal:

- the `-    ` indentation isn't natural when writing Markdown
- code examples should be carefully crafted for simplicity/readability and shouldn't necessarily follow the same code formatting as the actual code of the project

I don't think it's necessary to put too many rules on documentation.

package.json
```diff
@@ -1,84 +1,96 @@
...
+        "npm": ">=7.7.6"
+    },
+    "files": [
+        "/dist"
+    ],
+    "main": "dist/src/plugin.js",
+    "scripts": {
+        "lint:fix": "eslint --fix .",
+        "lint": "eslint .",
+        "check-format": "prettier --check .",
```
mnapoli, line 80: I'm not sure a new command is needed, we already have `npm run lint`.

In case I'm wrong, it might be worth updating https://github.com/getlift/lift/blob/master/.github/CONTRIBUTING.md

package.json
```diff
@@ -1,84 +1,96 @@
...
+        "type": "tsc --noEmit",
+        "build": "etsc",
+        "watch": "nodemon",
+        "test": "jest",
+        "prepare": "husky install"
```
mnapoli, line 85: I'm not a fan of precommit hooks, these things can be really surprising when not familiar with it.

Is this opt-in? Also curious: is this common practice in the node package ecosystem?

mnapoli (commented):
Hey, a few things:

Could you revert changes on ` package-lock.json` and exclude it? I feel like formatting it will be a nightmare of conflicts. This is an autogenerated file, let's not try to change it.

> a new test in the CI which ensure that all files are correctly formatted with prettier

I couldn't find that new job, did I miss it?

As for VS Code configuration, I'll let @fredericbarthelet review that to see if that makes sense to commit it (I'm not using it).

.eslintrc
adriencaccia: Yes, I only formatted the file with `prettier`

.github/CONTRIBUTING.md
adriencaccia: Done !

package.json
adriencaccia: `eslint` will only format js and ts files, for all other files, we need to use `prettier`.
It's a better practice to separate the two, since `eslint` is doing static analysis and `prettier` is only formatting.
That being said, I could change the existing commands to:

```json
    "lint:fix": "prettier --write . && eslint --fix .",
    "lint": "prettier --check . && eslint .",
```

What do you think ?

package.json
adriencaccia: With this setup, it is not opt-in no. The hook will be automatically installed when you run a `npm install` command.

In my experience, such hooks that automatically lint your code is a common practice in the node package ecosystem yes :slightly_smiling_face:

adriencaccia:
> Could you revert changes on ` package-lock.json` and exclude it? I feel like formatting it will be a nightmare of conflicts. This is an autogenerated file, let's not try to change it.

Done !
 
> > a new test in the CI which ensure that all files are correctly formatted with prettier
> 
> I couldn't find that new job, did I miss it?

I added it in one commit but removed it in another, I fixed it.

fredericbarthelet (approved):
Thanks @adriencaccia for making non JS files correctly formatted :) LGTM !

mnapoli:
Thanks @adriencaccia!

Issue #47, opened 2021-06-19
Permissions for queue construct


Pigius:
Hello, thank you for the development of the Lift plugin, it's so awesome, I hope it will support more use cases in the near future. I especially like it as a Ruby developer, as AWS CDK does not support it at all 😞 .

## Background:

Yesterday I've submitted the [PR](https://github.com/serverless/examples/pull/620) for the Serverless Framework example (SQS + DynamoDB). I've used the `queue construc`t for the example which is awesome.

However, during the implementation, I've crossed upon one problem, not sure whether is due to the construct, or maybe due to my lack of knowledge in some of the aspects of Serverless Framework.

## Problem

Within the  [permissions](https://github.com/getlift/lift/blob/master/docs/queue.md#permissions) section, we can see the information about:

> By default, all the Lambda functions deployed in the same serverless.yml file will be allowed to push messages into the queue.

> In the example below, there are no IAM permissions to set up: myFunction will be allowed to send messages into my-queue.

```yml
constructs:
    my-queue:
        type: queue
        # ...

functions:
    myFunction:
        handler: src/publisher.handler
        environment:
            QUEUE_URL: ${construct:my-queue.queueUrl}
```

This is quite nice, as I do not need to provide IAM statements within the `serverless.yml`.

However, when I've added another resource, which is DynamoDB in my case, and added the IAM statements:

```yml
iam:
    role:
      statements:
        - Effect: Allow
          Action:
            - dynamodb:PutItem
          Resource:
            - !GetAtt Table.Arn
 ```
After the deployment, the Lambda function for the `myFunction` will not have permissions to access the SQS anymore. So to make it work I need to add to the above iam role statements:
```yml
        - Effect: Allow
          Action:
             - sqs:SendMessage
          Resource: ${construct:lotteryQueue.queueArn}
```

Not sure, whether is a bug or not. 🤔

t-richard:
I think this is because Lift uses the now deprecated `provider.iamRoleStatements` to add permissions and Serverless framework ignores it if `provider.iam.role.statements` is set like in your case.

@Pigius Could you confirm using the following is working as expected ?

```
provider:
    iamRoleStatements:
        - Effect: Allow
          Action:
            - dynamodb:PutItem
          Resource:
            - !GetAtt Table.Arn
```

Pigius:
@t-richard Thanks for the fast reply. I've checked with the deprecated `provider.iamRoleStatements` and it working seamlessly without a role for the SQS queue.

[...]

PR #49, opened 2021-06-19
fix: allow to use the new provider.iam.role.statements

t-richard:
Should fix https://github.com/getlift/lift/issues/47

Here we add to `provider.iam.role.statements` if it exists or add to `provider.iamRoleStatements` if not (so that we are still backward-compatible with older versions of the serverless framework.

TODO
* [x] Test it 
* [x] Fix the type issue

### DIFF
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -200,10 +200,10 @@ class LiftPlugin {
             return;
         }
 
-        if (this.serverless.service.provider.iam?.role?.statements !== undefined) {
-            this.serverless.service.provider.iam.role.statements.push(...statements);
+        const role = this.serverless.service.provider.iam?.role;
 
-            return;
+        if (typeof role === "object" && "statements" in role) {
+            role.statements?.push(...statements);
         }
 
         this.serverless.service.provider.iamRoleStatements = this.serverless.service.provider.iamRoleStatements ?? [];
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -204,6 +204,8 @@ class LiftPlugin {
 
         if (typeof role === "object" && "statements" in role) {
             role.statements?.push(...statements);
+
+            return;
         }
 
         this.serverless.service.provider.iamRoleStatements = this.serverless.service.provider.iamRoleStatements ?? [];
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -199,6 +199,13 @@ class LiftPlugin {
         if (statements.length === 0) {
             return;
         }
+
+        if (this.serverless.service.provider.iam?.role?.statements !== undefined) {
+            this.serverless.service.provider.iam.role.statements.push(...statements);
+
+            return;
+        }
+
         this.serverless.service.provider.iamRoleStatements = this.serverless.service.provider.iamRoleStatements ?? [];
         this.serverless.service.provider.iamRoleStatements.push(...statements);
     }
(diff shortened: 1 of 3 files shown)


t-richard:
Not sure how to fix the CI. 

> Property 'statements' does not exist on type 'AwsLambdaRole | { name?: string | undefined; managedPolicies?: AwsArn[] | undefined; statements?: AwsIamPolicyStatements | undefined; permissionBoundary?: string | undefined; tags?: AwsResourceTags | undefined; }'
Property 'statements' does not exist on type 'string'.

It seems like the type definition is correct and that `statements` should match it. Not sure if the problem comes from the definition, the `tsc` checks or the code.

Let me know if you have any idea on how to fix.

mnapoli:
Thanks! The change itself looks really good. TypeScript doesn't approve of using/changing anything that isn't specified in the types (even if you double-check the property exists before).

I can take care of fixing the types on Monday! (and maybe add some tests too)

t-richard:
@mnapoli You're welcome.

I've fixed the type issue: basically, role can be a string, a cfn intrisinc expression or an object (which can contain `statements`). Ensuring we have an object filters out the string and checking that `statements` exists filters ou the intrisinc function and that it's not undefined. This way, Typescript is happy :smile: 

I've also added some tests but it's not covering all cases for now. If you think this is heading the right way, I'll check that again tomorrow.

test/unit/permissions.test.ts
```diff
@@ -0,0 +1,62 @@
+import { get, merge } from "lodash";
+import { pluginConfigExt, runServerless } from "../utils/runServerless";
+
+describe("permissions", () => {
+    it("should not override user-defined role", async () => {
```
mnapoli, line 5: 👌

test/unit/permissions.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    });
+
+    it("should append permissions when using the deprecated iamRoleStatements", async () => {
+        const { cfTemplate } = await runServerless({
+            fixture: "permissions",
+            configExt: merge(pluginConfigExt, {
+                provider: {
+                    iam: {
+                        role: {
+                            statements: [
```
mnapoli, line 31: The test name mentions `iamRoleStatements` but actually uses `iam.role.statements`, is this a mistake?

mnapoli:
> This way, Typescript is happy 😄

Perfect

> I've also added some tests but it's not covering all cases for now. If you think this is heading the right way, I'll check that again tomorrow.

And perfect again, yes those tests look good! I see you started that but ideally we should have:

- a test that covers `iamRoleStatements`
- a test that covers `iam.role.statements`

src/plugin.ts
```diff
@@ -199,6 +199,13 @@ class LiftPlugin {
         if (statements.length === 0) {
             return;
         }
+
+        const role = this.serverless.service.provider.iam?.role;
+
+        if (typeof role === "object" && "statements" in role) {
+            role.statements?.push(...statements);
```
fredericbarthelet, line 206: shouldn't you `return` here as well to prevent creation of `this.serverless.service.provider.iamRoleStatements` in case version of sls is up to date ?

t-richard: indeed, did it in my first commit but forgot it when refactoring to fix the type issue :+1:

test/unit/permissions.test.ts
t-richard: yes it is :sweat_smile:

t-richard:
Fixed the review comments and added some more tests to test with the new syntax, the old syntax and when no statements are provided.

fredericbarthelet (approved):

mnapoli:
Thank you!

PR #50, opened 2021-06-21
Instantiate constructs after all variables are resolved

### LINKED ISSUE #29: `${param}` variables are not resolved in constructs

mnapoli:
Fix #29.

Replaces #31.

On master:

- constructs are instantiated in the constructor of the Lift plugin
- that means that variables resolved _after_ the Lift plugin is instantiated are not resolved in constructs

In this branch:

- constructs are instantiated in the `initialize` hook (at this point, all variables are resolved)
- to achieve that, commands are now defined statically
- and Lift variables are resolved lazily using CDK tokens (CDK lazy variables)

This PR includes #41 (the test that reproduces the problem).

TODO:

- [x] additional testing

### DIFF
--- a/test/fixtures/variables/extra-plugin.js
+++ b/test/fixtures/variables/extra-plugin.js
@@ -20,6 +20,17 @@ module.exports = class ExtraPlugin {
                     };
                 },
             },
+            "custom-arn": {
+                // Async resolver replicating the Dashboard Plugin implementation
+                // https://github.com/serverless/dashboard-plugin/blob/ea662fcf03e8f2f5c1f435d24249c72025a24dbf/lib/plugin.js#L180-L194
+                async resolve() {
+                    return {
+                        value: await (async () => {
+                            return Promise.resolve("arn:aws:acm:us-east-1:123466615250:certificate/abcdef-b896-4725-96e3-6f143d06ac0b");
+                        })(),
+                    };
+                },
+            },
         };
         // Old variable mechanism
         this.variableResolvers = {
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,3 +2,4 @@
 /lib
 /dist
 /utils
+/test/fixtures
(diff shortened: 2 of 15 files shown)


test/fixtures/variables/extra-plugin.js
```diff
@@ -0,0 +1,45 @@
+module.exports = class ExtraPlugin {
```
fredericbarthelet, line 1: Why not in TS ? This would allow you not to have to ignore this file in `.eslintignore`

fredericbarthelet (approved):

test/fixtures/variables/extra-plugin.js
mnapoli: Good point I did not think about that.

However I think that might bring in more complexity: I would have to include the compiled `*.js` plugin file (not the `*.ts` one). Also I don't really want to bother defining types everywhere: this is a fixture file and it's replicating existing JS code that I copy/pasted from problematic scenarios from other JS plugins so…

PR #51, opened 2021-06-21
[Static Website] Add new `errorPage` configuration

mnapoli:
At the moment, `static-website` redirects all 404 to `index.html`. This is perfect for SPA with routing in the browser (e.g. with react router or vue router): we want all unknown URLs to load the JS app.

However, this doesn't really work with flat static websites (without client-side routing).

This PR introduces a new `errorPage` option:

```yaml
constructs:
    landing:
        type: static-website
        # ...
        errorPage: error.html
```

- when not set, `static-website` is optimized for SPA and redirects 404s to `index.html` with a 200 response
- when set, 404s serve the error page (and keep a 404 response)

> Why go with the "SPA" behavior by default? Why not respond `404` responses by default?

The default error response for S3 is extremely ugly:

image

This cannot be considered a good "production-ready default". So the default config could be a choice between:

1. serving a default error HTML page provided by Lift (like a 404 Netlify page)
2. enable the SPA behavior by default

In case of 1, that means that users building SPA would have to enable a new option to get SPA behavior -> I think this could be the most popular use case, compared to flat file websites. That's why I think enabling the SPA behavior by default makes sense.

After all, anyone building a flat-file static website will want to customize the error page anyway, right?

### DIFF
--- a/docs/static-website.md
+++ b/docs/static-website.md
@@ -162,6 +162,23 @@ constructs:
             - app.mywebsite.com
 ```
 
+### Error page
+
+By default, all 404 requests are redirected to `index.html` with a 200 response status. This behavior is optimized for Single-Page Applications: it allows doing client-side URL routing with JavaScript frameworks.
+
+For static websites that _are not SPA_, it is possible to serve a custom "Not found" error page:
+
+```yaml
+constructs:
+    landing:
+        # ...
+        errorPage: error.html # can be any HTML file in your project
+```
+
+When a browser requests the URL of a non-existing file, the `error.html` file will be served with a 404 response status.
+
+Do not use this setting when doing JavaScript URL routing: this will break URL routing.
+
 ### Allow iframes
 
 By default, as recommended [for security reasons](https://scotthelme.co.uk/hardening-your-http-response-headers/#x-frame-options), the static website cannot be embedded in an iframe.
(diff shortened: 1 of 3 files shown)


docs/static-website.md
```diff
@@ -162,6 +162,25 @@ constructs:
...
+    landing:
+        # ...
+        errorPage: error.html # can be any HTML file in your project
+```
+
+When a browser requests the URL of a non-existing file, the `error.html` file will be served with a 404 response status.
+
+Do not use this setting when doing JavaScript URL routing: this will break URL routing.
+
+When
```
t-richard, line 182: Something is wrong here

src/constructs/StaticWebsite.ts
```diff
@@ -278,6 +272,34 @@ export class StaticWebsite extends AwsConstruct {
         return this.provider.getStackOutput(this.distributionIdOutput);
     }
 
+    private errorResponse(): ErrorResponse {
+        // Custom error page
+        if (this.configuration.errorPage !== undefined) {
+            let errorPath = this.configuration.errorPage;
+            if (!errorPath.startsWith("/")) {
+                errorPath = `/${errorPath}`;
```
t-richard, line 280: If errorPath begins with a dot like `./error.html` then it would be invalid.

I think this case should be covered.

mnapoli: Good point, I have added a check for that.

t-richard (approved):

PR #52, opened 2021-06-21
[RFC] Deploy any CDK construct | v2

mnapoli:
This PR replaces #23

This PR is based on #50 (the diff is more complex than it actually is)

**Note: this is a Request For Comments, not meant to be merged right now.**

Instead of creating a Lift construct that allows importing any CDK construct, this PR brings support for 3rd party **Lift constructs**.

This allows developers to use any CDK construct, **but also to add Lift features**, like:

- construct outputs
- construct variables
- automatic IAM permissions
- custom commands
- etc.

## Example

```yaml
service: app

provider:
    name: aws

constructs:
    cron:
        type: ./construct
        # any custom option

plugins:
    - serverless-lift
```

☝️  you may notice that the construct type points to the JavaScript file that contains the Lift construct. That's how discovery happens.

Example of a construct file (`construct.js`):

```js
const events = require('@aws-cdk/aws-events');
const targets = require('@aws-cdk/aws-events-targets');
const lambda = require('@aws-cdk/aws-lambda');
const cdk = require('@aws-cdk/core');

const fs = require('fs');

class Cron extends cdk.Construct {
    // JSON schema to validate the construct's configuration in serverless.yml
    static schema = {
        type: "object",
        properties: {
            timeout: { type: "number" },
        },
        additionalProperties: false,
        required: [],
    };

    static create(provider, id, configuration) {
        return new this(provider.stack, id, configuration);
    }

    constructor(app, id, configuration) {
        super(app, id);

        const lambdaFn = new lambda.Function(this, 'Handler', {
            code: new lambda.InlineCode(fs.readFileSync('lambda-handler.py', { encoding: 'utf-8' })),
            handler: 'index.main',
            timeout: cdk.Duration.seconds(3),
            runtime: lambda.Runtime.PYTHON_3_6,
        });

        const rule = new events.Rule(this, 'Rule', {
            schedule: events.Schedule.expression('rate(5 minutes)')
        });

        rule.addTarget(new targets.LambdaFunction(lambdaFn));
    }
}

module.exports = Cron
```

A construct can take advantage of Lift features, for example exposing CLI outputs:

```js
class Cron extends cdk.Construct {
    constructor(app, id, configuration, provider) {
        # ...

        this.provider = provider;

        this.functionNameOutput = new CfnOutput(this, "HandlerFunctionName", {
            value: lambdaFn.functionName,
        });
    }

    /**
     * CLI outputs
     */
    outputs() {
        return {
            functionName: () => this.provider.getStackOutput(this.functionNameOutput),
        };
    }
}

module.exports = Cron
```

Lift constructs written in TypeScript should implement the `ConstructInterface` and `StaticConstructInterface`: https://github.com/getlift/lift/blob/9977a8ac7c447953c403fea74e4ce51cbf7cb4a2/src/classes/Construct.ts#L8-L43

In order to ease implementing constructs, we may also consider making most fields optional (e.g. defining a JSON schema, etc.). Most of them are already anyway.

### DIFF
--- a/src/classes/Construct.ts
+++ b/src/classes/Construct.ts
@@ -1,5 +1,6 @@
 import { PolicyStatement } from "../CloudFormation";
 import { AwsProvider } from "./AwsProvider";
+import ServerlessError from "../utils/error";
 import { CliOptions } from "../types/serverless";
 
 /**
@@ -55,3 +56,37 @@ type ConstructCommandDefinition = {
         };
     };
 };
+
+export function assertValidConstructClass(constructClass: unknown): asserts constructClass is StaticConstructInterface {
+    if (typeof constructClass !== "function") {
+        throw new ServerlessError(
+            `Tried to register a construct that is not an object. The construct is of type '${typeof constructClass}'`,
+            "LIFT_CONSTRUCT_INVALID_OBJECT"
+        );
+    }
+    // Type
+    if (!hasProperty(constructClass, "type") || typeof constructClass.type !== "string") {
+        throw new ServerlessError(
+            `The construct '${constructClass.constructor.name}' does not expose a static 'type' string property. ` +
+                "All constructs must expose a type name via a 'type' property.",
+            "LIFT_CONSTRUCT_MISSING_TYPE"
+        );
+    }
+    // Schema
+    if (!hasProperty(constructClass, "schema") || typeof constructClass.schema !== "object") {
+        throw new ServerlessError(
+            `The construct '${constructClass.type}' does not expose a static 'schema' property. ` +
+                "All constructs must expose a JSON schema via a 'schema' property.",
+            "LIFT_CONSTRUCT_MISSING_SCHEMA"
+        );
+    }
+}
+
+/**
+ * This crazy function only exists to make TypeScript happy with type narrowing
+ */
+// eslint-disable-next-line @typescript-eslint/ban-types
+function hasProperty<X extends {}, Y extends PropertyKey>(obj: X, prop: Y): obj is X & Record<Y, unknown> {
+    // eslint-disable-next-line no-prototype-builtins
+    return obj.hasOwnProperty(prop);
+}
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -5,6 +5,7 @@ import * as path from "path";
 import { readFileSync } from "fs";
 import { dump } from "js-yaml";
 import { DefaultTokenResolver, Lazy, StringConcat, Tokenization } from "@aws-cdk/core";
+import { set } from "lodash";
 import type {
     CommandsDefinition,
     DeprecatedVariableResolver,
@@ -14,7 +15,7 @@ import type {
 } from "./types/serverless";
 import { AwsProvider, ConstructInterface } from "./classes";
 import { log } from "./utils/logger";
-import { StaticConstructInterface } from "./classes/Construct";
+import { assertValidConstructClass, StaticConstructInterface } from "./classes/Construct";
 import ServerlessError from "./utils/error";
 
 const CONSTRUCT_ID_PATTERN = "^[a-zA-Z0-9-_]+$";
@@ -91,6 +92,7 @@ class LiftPlugin {
         };
 
         this.registerProviders();
+        this.loadCustomConstructs();
         this.registerConstructsSchema();
         this.registerConfigSchema();
         this.registerCommands();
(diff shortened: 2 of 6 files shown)


t-richard:
This approach looks way better to me than V1.

The fact that you can give to your constructs "lift superpowers" like variables makes it totally awesome.

I like the syntax too, it's clear and simple 👌 

One thing though. When using plain javascript it is  quite straight forward. 

But what if I want to use typescript? I guess I should point to the compiled file in the typescript output directory? How would it play with serverless typescript plugin or serverless webpack plugin? 

Not that it needs to be integrated now, but it's worth mentioning in the docs if it's compatible and how to use this with typescript IMHO.

mnapoli:
> But what if I want to use typescript?

Yes those are good points, I think you'd have to point to the compiled file indeed, but I have no idea how that would work with those plugins yet.

I think the DX is much better with TypeScript, so that would be something to know about/document eventually.

t-richard:
The nice thing with the serverless-typescript-plugin is that you can reference Typescript files directly and the plugin will take care of compiling before deploy and resolve filenames to the compiled ones in your function declaration.

See this example repo https://github.com/prisma-labs/serverless-plugin-typescript/tree/424fe4906bd15ceeeda6139dfe9e0f8bf873c4c3/example

Here `handler.ts` would be resolved to something like `dist/handler.js`. 

I'm also wondering if that would work for queues (because you are defining the handler in the queue construct directly, not sure if the plugin would be able to see it).

mnapoli:
OK I gave it a try but serverless-typescript-plugin doesn't help writing TS constructs: these are compiled after Lift boots and tries to load the constructs.

However I had a look at how the plugin works, it is very likely that TS queue functions should work with that plugin.

t-richard:
> these are compiled after Lift boots and tries to load the constructs

So you would have to compile beforehand ? I think that plugins order matters, did you added the typescript-plugin before Lift ?
Not sure it's relevant here, but it is when using the plugin with serverless-offline for example.

See https://github.com/prisma-labs/serverless-plugin-typescript/tree/424fe4906bd15ceeeda6139dfe9e0f8bf873c4c3#usage-with-serverless-offline

mnapoli:
Serverless Framework supports using `serverless.ts` instead of `serverless.yml`. They do so via `ts-node`, a JIT TypeScript parser:

https://github.com/serverless/serverless/blob/f16c45f84b8be1e469bfdd92191fc760b8f1631e/lib/configuration/read.js#L111-L136

We could do the same here:

- either require `ts-node`
- or go the Serverless Framework route: expect it to be installed if users want to use Typescript

TastefulElk:
Hey guys, really appreciate the effort put in here, you guys rock! 🎸 Just thought I'd chip in with some thoughts if you don't mind!

It's nice to be able to define constructs yourself but, in my opinion, if I have to do it in the same project, I might as well write it in CloudFormation. What would be **really** useful however is if I could take that supercharged Lift construct, package it up, and then re-use it with something like

```yml
constructs:
    cron:
        type: @lift/some-construct
        # any custom option
```

That would let this project scale with the help of the community on a whole other level. 

I _think_ that would be the primary way custom constructs would be used, rendering the local construct TS transpilation issue almost an edge case for which requiring `ts-node` to pre-exist would be fine.

TastefulElk:
@mnapoli I mean package it up as in creating an npm package out of it so that I can re-use the same custom construct in multiple projects! 👍

mnapoli:
oh 🤦  yes! This is definitely what we are aiming for eventually!

walmsles:
+1 to @TastefulElk 's suggestion above!  This is exactly my use case, the ability to use constructs from an installed npm package that a central team can control and administer.  Allows for great re-use and fits our existing use of the Serverless Framework for delivery of serverless solutions.

walmsles:
Just referencing my exact use case from the other issue : https://github.com/getlift/lift/issues/104#issuecomment-932156990

@mnapoli 
Other question around Lift:  Current implementation is through providers and AWS only (which makes sense right now), is the goal to also provide a similar construct idea available to other cloud providers within the Serverless framework?

mnapoli:
@walmsles we are indeed considering that idea. We're looking into supporting constructs for, e.g. FaunaDB, Netlify websites, Cloudflare workers, Algolia, etc. instead of Azure & GCP (which are already supported by a Serverless Framework plugin).

If you have any need or use case around that I'm really interested to hear it.

BrutalSimplicity:
I wish custom constructs were the main driver here, and this was exposed by default. It seems there are just too many edge cases to solve for everyone with the constructs included. This project on its own with an exposed interface for creating custom constructs would have been more than enough.

mnapoli:
@BrutalSimplicity thanks for the feedback. I know it's been said before, but we're actively looking into that.

mnapoli:
Hey everyone, I'm reviving this topic: if you were able to deploy CDK constructs in Serverless Framework, do you have an immediate use case for it?

I.e. instead of "this might be great/interesting theoretically", do you have specific examples of resources/CDK constructs you would try to deploy as soon as this was available?

piolet:
Hi Matthieu, 

you know my main use case ;)

A server side site able to handle also static pages in assets ;)

But for a specific resource, i think an ECS to deploy an quickly a docker image.
ideally, attach this ecs to a load balancer to allow access via a domain name

aldebout:
@mnapoli this is the construct I would use tomorrow if this was available :) https://docs.aws.amazon.com/solutions/latest/constructs/aws-eventbridge-kinesisfirehose-s3.html

fredericbarthelet:
Discussed in https://github.com/getlift/lift/discussions/182. Please use this discussion for further implementation of the feature.

PR #53, opened 2021-06-22
Add npm prepare script

t-richard:
This is PR adds a `prepare` script to the package.json.

This is a special npm script that is executed when installing a package from a git URL like so

```bash
npm install git@github.com:t-richard/lift.git#add-prepare-script
```

This would come in handy for the community to test WIP/draft constructs and is much more user-friendly than using `npm link`.

Previously, doing this would not work because we would be missing the `dist` folder in ` node_modules/serverless-lift`

The script behaviour is described at https://docs.npmjs.com/cli/v7/using-npm/scripts#life-cycle-scripts

It also says
> Runs BEFORE the package is published

This is why I also deleted `npm run build` from the release GH Action workflow

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,8 @@
     "type": "tsc --noEmit",
     "build": "etsc",
     "watch": "nodemon",
-    "test": "jest"
+    "test": "jest",
+    "prepare": "etsc"
   },
   "types": "lib/index.d.ts"
 }
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "build": "etsc",
     "watch": "nodemon",
     "test": "jest",
-    "prepare": "etsc"
+    "prepare": "npm run build"
   },
   "types": "lib/index.d.ts"
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -78,7 +78,8 @@
     "type": "tsc --noEmit",
     "build": "etsc",
     "watch": "nodemon",
-    "test": "jest"
+    "test": "jest",
+    "prepare": "etsc"
```
fredericbarthelet, line 82:
```suggestion
    "prepare": "build"
```

This way, if we'd change in the future the build step for the project, `prepare` would be updated as well :)

fredericbarthelet (changes requested):
Looks  like a nice improvement :) ! Thanks @t-richard !
I suggested a minor improvement to ensure coherence between build and prepare. Let me know what you think of it.

t-richard:
@fredericbarthelet that makes sense. I wasn't able to make it work with your suggestion. 

Just using "build" is not enough apparently, all solutions I found online suggested to use the full command like `npm run build` so this is what I did and it works as expected.

fredericbarthelet (approved):
LGTM. Thanks for your contribution @t-richard :) !

PR #58, opened 2021-06-25
fix addFunction to support arrays

bboure:
The Serverless Framework supports passing an [array of functions](https://www.serverless.com/framework/docs/providers/aws/guide/functions/).

When `addFunction` is called, it assumes it is an object. As a result, the function is not added to the CloudFormation template.

### DIFF
--- a/src/classes/AwsProvider.ts
+++ b/src/classes/AwsProvider.ts
@@ -72,9 +72,13 @@ export class AwsProvider {
             this.serverless.configurationInput.functions = {};
         }
 
-        Object.assign(this.serverless.configurationInput.functions, {
-            [functionName]: functionConfig,
-        });
+        if (Array.isArray(this.serverless.configurationInput.functions)) {
+            this.serverless.configurationInput.functions.push({ [functionName]: functionConfig });
+        } else {
+            Object.assign(this.serverless.configurationInput.functions, {
+                [functionName]: functionConfig,
+            });
+        }
     }
 
     /**


t-richard:
@bboure Just to be sure, you are referring to this syntax right ?

```yaml
functions:
  - ${file(../foo-functions.yml)}
  - ${file(../bar-functions.yml)}
```

bboure:
@t-richard Correct!

t-richard (approved):

mnapoli:
Hey, since merging #50 there were conflicts that I've resolved in https://github.com/getlift/lift/compare/master...fix/addFunction

However, as I added some tests to cover this, I think the bug disappeared with #50. Indeed, we no longer modify `serverless.configurationInput.functions`: now we add functions to `serverless.service.functions`. And it seems I cannot get this bug to reproduce now.

Could you share with me an example of the case where you got the bug?

You can see the test case that I tried writing: https://github.com/getlift/lift/compare/master...fix/addFunction#diff-d34fa05fa3de68cc5761ced953a68777b838e78a0e91a35ab4fa4309c059a4f1R9-R16 This did not cause any problem (when I remove your fix).

bboure:
Thanks @mnapoli 

My use case was exactly what @t-richard said
```
functions:
  - ${file(../foo-functions.yml)}
  - ${file(../bar-functions.yml)}
```

I just tested on the master branch and the bug is gone with the latest changes.

mnapoli:
👍  thank you for following up!

PR #59, opened 2021-06-28
Add matrix strategy

fredericbarthelet:
This PR implements the following features :

1. Add multiple runs of Jest unit test with node v10, V12, v14 and v15
2. Add multiple runs of Jest unit test with serverless v2.36.0 and latest
3. `Lint` and `Type` tests were moved out of the testing matrix as they are not impacted by node or serverless version

### Modifications made to implement 1

In order to allow installing and running of tests for older version of node, package-lock cannot be used.
`package-lock.json` was not distributed as part of the npm package and using one in dev environment [can prevent noticing problems actually encountered by lib users](https://www.twilio.com/blog/lockfiles-nodejs).

- `package-lock.json` was removed and option `package-lock=false` was added to `.npmrc`
- `engines` version for node were opened from node 10 to node 15

### Modifications made to implement 2

Previously minimum version of serverless required was `2.35.0`.
However, current serverless version `2.49.0` uses `@serverless/test@8` that breaks backward compatibility of our test suite. This change was introduced in `2.36.0`. This version is now registered in `peerDependencies` section of package.json. This plugin might work for older versions of serverless, but no test suite can be implemented in this package to guarantee a flawless experience.

- updating to latest `prettier` version actually brought additional format fixes that are implemented in this PR
- serverless error message changed from version `2.36.0` to `2.49.0`. Tests that checked equality of Error string were refactored to use a regular expression rather than a strict equality, allowing both error message to pass the test.


.github/workflows/ci.yml
```diff
@@ -6,34 +6,31 @@ on:
...
+    test:
         name: Build
         runs-on: ubuntu-latest
+        strategy:
+            matrix:
+                node:
+                    # - 10
+                    # - 12
+                    # - 14
+                    - 15
```
mnapoli, line 18: I see that at the moment other Node versions are commented, I'm just opening the discussion in advance: do we need to test with Node 12 and 14? Or can we target the lowest/highest (10/15) and be confident that it should work with the rest?

.github/workflows/ci.yml
```diff
@@ -6,34 +6,31 @@ on:
     pull_request:
 
 jobs:
-    build:
+    test:
         name: Build
```
mnapoli, line 10:
```suggestion
        name: Tests
```

.github/workflows/ci.yml
fredericbarthelet: All tests run in parallel, there is at the moment no additional cost brought by actually testing all node major between 10 and 15. Therefor I do not see any reason not to request it as part of the testing matrix :)

Issue #63, opened 2021-06-30
DLQ URL/ARN not available


M1ke:
When trying to use a DLQ in another component via:

```
destinations:
      onFailure: ${construct:queue.dlqUrlOutput}
```

I am getting the error:

```
Resolution error: ${construct:queue.dlqUrlOutput} does not exist. Properties available on ${construct:queue} are: queueUrl, queueArn.
```

However the below linked line seems to suggest this does get output:

https://github.com/getlift/lift/blob/25192e628a1ed25c2b1beb5866c289c1237f2341/src/constructs/Queue.ts#L140

M1ke:
Actually having re-read this even if it was available it wouldn't work as I'd need the ARN. But either way, this looks like both should be available.

t-richard:
When looking for available variables, you have too look at `references`

https://github.com/getlift/lift/blob/25192e628a1ed25c2b1beb5866c289c1237f2341/src/constructs/Queue.ts#L154

Outputs are just the Cloudformation stack outputs but are not directly related to what's available in Lift variables.

But yeah, that could be cool to have those too.

What is the use case here ?

mnapoli:
Hey, I think your use case makes sense, but I'm not sure it should be solved by reusing the DLQ from the queue component to be honest.

To me, it seems like you rather need a DLQ on the Lambda handling eventbridge messages, right?

Maybe it would make sense to have a (production-ready) "event handler" construct for handling eventbridge messages? That construct could include all the default config for production scenarios, like retries, DLQ, etc.

Here is an illustration:

Before:

```yaml
provider:
  ...
  eventBridge:
    useCloudFormation: true

functions:
  myfunc:
    handler: index.handler
    events:
      - eventBridge:
          eventBus: ...
          pattern: ...
          inputTransformer: ...
      - eventBridge:
          eventBus: ...
          pattern: ...
          inputTransformer: ...
    destinations:
      onFailure: ${construct:queue.dlqUrlOutput}
```

After:

```yaml
constructs:

  myfunc:
    type: event-handler
    handler: index.handler
    events:
      # only eventbridge events allowed here
      - eventBridge:
          eventBus: ...
          pattern: ...
          inputTransformer: ...
      - eventBridge:
          eventBus: ...
          pattern: ...
          inputTransformer: ...
```

That will automatically configure a SQS dead letter queue for this handler.

Just like with queues, the construct could provide the commands:

```sh
serverless myfunc:failed
# list failed messages from the dlq

[...]

PR #64, opened 2021-06-30
Add DLQ URL & ARN to outputs

M1ke:
Resolves https://github.com/getlift/lift/issues/63

This may be a naive addition because I can see that the format of the references for `this.queue` is different and possibly it's not as easy to replicate. But it's a start

### DIFF
--- a/src/constructs/Queue.ts
+++ b/src/constructs/Queue.ts
@@ -61,6 +61,7 @@ export class Queue extends AwsConstruct {
     private readonly queueArnOutput: CfnOutput;
     private readonly queueUrlOutput: CfnOutput;
     private readonly dlqUrlOutput: CfnOutput;
+    private readonly dlqArnOutput: CfnOutput;
 
     constructor(
         scope: CdkConstruct,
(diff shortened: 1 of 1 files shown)


src/constructs/Queue.ts
```diff
@@ -155,6 +160,8 @@ export class Queue extends AwsConstruct {
         return {
             queueUrl: this.referenceQueueUrl(),
             queueArn: this.referenceQueueArn(),
+            dlqUrl: this.getDlqUrl(),
+            dlqArn: thus.getDlqArn(),
```
t-richard, line 164: you can't use those functions as they use the Cloudformation stack outputs. But you need the references (Cloudformation's `Ref` or `GetAtt`) to be able to reference it custom cloudformation, function environment variables, event triggers on function etc

That's the purpose of `AwsProvider::getCloudFormationReference`. This function will use the CDK to resolve references which in the end uses CDK Tokens IIRC. 

Check `referenceQueueUrl` to get an idea of how to do it here.

t-richard, line 164: and there is a type too. `thus` instead of `this`

src/constructs/Queue.ts
```diff
@@ -197,6 +204,10 @@ export class Queue extends AwsConstruct {
         return this.provider.getStackOutput(this.dlqUrlOutput);
     }
 
+    async getDlqArn(): Promise<string | undefined> {
```
t-richard, line 207: I think this is not needed regarding my previous comment.

t-richard (changes requested):
Thanks for starting this.

Not sure if this is needed and wanted in Lift to be honest. But maybe if you need it, some others will too.

So here are some comments on the implementation.

mnapoli:
Following the discussion in https://github.com/M1ke/serverless-payments/pull/8 I don't think we should merge this.

I'll close this for now, we can always reopen later!

Thanks for the time spent on this!

PR #65, opened 2021-07-01
[RFC] VPC construct

t-richard:
This is a draft PR related to the VPC RFC.

This implementation is far from complete and is here more to open a discussion than to be merged immediatly.

TODO

- [x] Add proper tests
- [ ] ~~support more options~~
- [ ] ~~add references~~
- [ ] ~~add outputs~~
- [ ] documentation
- [x] properly handle account/regions/AZs (only support 2AZ for now so we are in an environment-agnostic setup)

### DIFF
--- a/src/classes/AwsConstruct.ts
+++ b/src/classes/AwsConstruct.ts
@@ -1,4 +1,5 @@
 import { Construct as CdkConstruct } from "@aws-cdk/core";
+import { AwsCfInstruction } from "@serverless/typescript";
 import { ConstructInterface } from ".";
 import { AwsProvider } from "./AwsProvider";
 
@@ -29,5 +30,5 @@ export abstract class AwsConstruct extends CdkConstruct implements ConstructInte
     /**
      * CloudFormation references
      */
-    abstract references(): Record<string, Record<string, unknown>>;
+    abstract references(): Record<string, AwsCfInstruction>;
 }
--- a/src/classes/AwsProvider.ts
+++ b/src/classes/AwsProvider.ts
@@ -88,7 +88,7 @@ export class AwsProvider {
     /**
      * @internal
      */
-    setVpcConfig(securityGroup: AwsCfInstruction, subnets: AwsCfInstruction[]): void {
+    setVpcConfig(securityGroups: AwsCfInstruction[], subnets: AwsCfInstruction[]): void {
         if (this.getVpcConfig() !== null) {
             throw new ServerlessError(
                 "Can't register more than one VPC.\n" +
@@ -99,7 +99,7 @@ export class AwsProvider {
         }
 
         this.serverless.service.provider.vpc = {
-            securityGroupIds: [securityGroup], // TODO : merge with existing groups ?
+            securityGroupIds: securityGroups, // TODO : merge with existing groups ?
             subnetIds: subnets,
         };
     }
--- a/src/classes/AwsProvider.ts
+++ b/src/classes/AwsProvider.ts
@@ -1,5 +1,6 @@
 import { App, CfnOutput, Stack } from "@aws-cdk/core";
 import { get, merge } from "lodash";
+import { AwsCfInstruction, AwsLambdaVpcConfig } from "@serverless/typescript";
 import { getStackOutput } from "../CloudFormation";
 import { CloudformationTemplate, Provider as LegacyAwsProvider, Serverless } from "../types/serverless";
 import { awsRequest } from "./aws";
@@ -8,6 +9,7 @@ import { StaticConstructInterface } from "./Construct";
 import ServerlessError from "../utils/error";
 import { Storage } from "../constructs/Storage";
 import { Queue } from "../constructs/Queue";
+import { VPC } from "../constructs/VPC";
 import { Webhook } from "../constructs/Webhook";
 import { StaticWebsite } from "../constructs/StaticWebsite";
 
@@ -42,14 +44,13 @@ export class AwsProvider {
     public naming: { getStackName: () => string; getLambdaLogicalId: (functionName: string) => string };
 
     constructor(private readonly serverless: Serverless) {
+        this.stackName = serverless.getProvider("aws").naming.getStackName();
         this.app = new App();
         this.stack = new Stack(this.app);
-        serverless.stack = this.stack;
-        this.stackName = serverless.getProvider("aws").naming.getStackName();
-
         this.legacyProvider = serverless.getProvider("aws");
         this.naming = this.legacyProvider.naming;
         this.region = serverless.getProvider("aws").getRegion();
+        serverless.stack = this.stack;
     }
 
     create(type: string, id: string): ConstructInterface {
@@ -85,6 +86,35 @@ export class AwsProvider {
     }
 
     /**
+     * @internal
+     */
+    setVpcConfig(securityGroup: AwsCfInstruction, subnets: AwsCfInstruction[]): void {
+        if (this.getVpcConfig() !== null) {
+            throw new ServerlessError(
+                "Can't register more than one VPC.\n" +
+                    'Either you have several "vpc" constructs \n' +
+                    'or you already defined "provider.vpc" in serverless.yml',
+                "LIFT_ONLY_ONE_VPC"
+            );
+        }
+
+        this.serverless.service.provider.vpc = {
+            securityGroupIds: [securityGroup], // TODO : merge with existing groups ?
+            subnetIds: subnets,
+        };
+    }
+
+    /**
+     * This function can be used by other constructs to reference
+     * global subnets or security groups in their resources
+     *
+     * @internal
+     */
+    getVpcConfig(): AwsLambdaVpcConfig | null {
+        return this.serverless.service.provider.vpc ?? null;
+    }
+
+    /**
      * Resolves the value of a CloudFormation stack output.
      */
     async getStackOutput(output: CfnOutput): Promise<string | undefined> {
@@ -94,8 +124,8 @@ export class AwsProvider {
     /**
      * Returns a CloudFormation intrinsic function, like Fn::Ref, GetAtt, etc.
      */
-    getCloudFormationReference(value: string): Record<string, unknown> {
-        return Stack.of(this.stack).resolve(value) as Record<string, unknown>;
+    getCloudFormationReference(value: string): AwsCfInstruction {
+        return Stack.of(this.stack).resolve(value) as AwsCfInstruction;
     }
 
     /**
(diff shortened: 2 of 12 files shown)


mnapoli:
Hey @t-richard, turns out I already have a WIP of a VPC construct 😅  It looks very similar btw, and it will go hand in hand with an RDS construct.

I just want to let you know before you invest much more work into this PR.

mnapoli:
Following discussions in the RFC and with Frédéric: let's go with your solution (having a proper "VPC" construct). Don't bother rebasing/incorporating my branch.

In order to start with a limited scope:

- it should be production-ready: 2 AZ, 1 NAT per AZ
- caring about a "dev" mode is out of scope for now
- because of that, let's ignore making the gateway type customizable for now (NAT Gateway only)

And more generally: don't add anything we don't have a use case for. That includes variables, outputs, etc. The reason is that it's easier to add anything later, but it's extremely hard to remove anything.

Let me know if I can help in any way!

src/classes/AwsProvider.ts
```diff
@@ -84,6 +88,17 @@ export class AwsProvider {
         this.serverless.service.setFunctionNames(this.serverless.processedInput.options);
     }
 
+    setVpcConfig(securityGroup: AwsCfInstruction, subnets: AwsCfInstruction[]): void {
```
mnapoli, line 91: This should probably error if there's already a VPC setup?

src/constructs/VPC.ts
```diff
@@ -0,0 +1,76 @@
...
+import { AwsConstruct, AwsProvider } from "../classes";
+
+const STORAGE_DEFINITION = {
+    type: "object",
+    properties: {
+        type: { const: "vpc" },
+        natGatewayType: {
+            anyOf: [{ const: "gateway" }, { const: "instance" }],
+        },
+        natGateways: { type: "number" },
```
mnapoli, line 22: I think we can remove those 2 settings for now.

test/unit/vpc.test.ts
```diff
@@ -0,0 +1,25 @@
+import { get } from "lodash";
+import { pluginConfigExt, runServerless } from "../utils/runServerless";
+
+type CfTemplate = {
+    Resources: Record<string, unknown>;
+    Outputs?: Record<string, unknown>;
+};
+
+function expectUserDynamoStatementIsAdded(cfTemplate: CfTemplate) {
```
mnapoli, line 9: Is "Dynamo" a leftover from another branch?

t-richard: yep, definitly copy/pasted some test code and did not cleaned-it-up :sweat_smile:

t-richard:
> In order to start with a limited scope:
>
> it should be production-ready: 2 AZ, 1 NAT per AZ
caring about a "dev" mode is out of scope for now
because of that, let's ignore making the gateway type customizable for now (NAT Gateway only)

That makes sense, thanks for the feedback. 

I'll try to work on that this week. This is almost a working solution for the described use case but it needs a bit or rework, more tests and things like that.

src/classes/AwsProvider.ts
t-richard: Yes, it should. And we could handle the "2 VPC constructs" case here too.

src/classes/AwsProvider.ts
```diff
@@ -84,6 +85,35 @@ export class AwsProvider {
         this.serverless.service.setFunctionNames(this.serverless.processedInput.options);
     }
 
+    /**
+     * @internal
```
mnapoli, line 89: Good idea, better err on the safe side first.

src/classes/AwsProvider.ts
```diff
@@ -84,6 +85,35 @@ export class AwsProvider {
...
+            throw new ServerlessError(
+                "Can't register more than one VPC.\n" +
+                    'Either you have several "vpc" constructs \n' +
+                    'or you already defined "provider.vpc" in serverless.yml',
+                "LIFT_ONLY_ONE_VPC"
+            );
+        }
+
+        this.serverless.service.provider.vpc = {
+            securityGroupIds: [securityGroup], // TODO : merge with existing groups ?
```
mnapoli, line 102: I think the TODO can be removed now? (if VPC settings are defined in serverless.yml then there will be an error)

src/classes/AwsConstruct.ts
```diff
@@ -29,5 +30,5 @@ export abstract class AwsConstruct extends CdkConstruct implements ConstructInte
     /**
      * CloudFormation references
      */
-    abstract references(): Record<string, Record<string, unknown>>;
+    abstract references(): Record<string, AwsCfInstruction>;
```
mnapoli, line 33: These changes have been merged on master, it looks like a rebase problem?

src/constructs/VPC.ts
```diff
@@ -0,0 +1,52 @@
...
+    constructor(scope: CdkConstruct, id: string, configuration: Configuration, private provider: AwsProvider) {
+        super(scope, id);
+
+        this.vpc = new Vpc(this, "VPC", {
+            maxAzs: 2,
+        });
+
+        const privateSubnets = this.vpc.selectSubnets({
+            subnetType: SubnetType.PRIVATE,
+        }).subnetIds;
```
mnapoli, line 33: There is the `vpc.privateSubnets` property, is there any difference? (I'm just arguing for `vpc.privateSubnets` because it is simpler)

Mmh after looking up, and unless I'm mistaken, it seems to be equivalent: https://github.com/aws/aws-cdk/blob/283ed02c64f827161edba3e11c3cead3b54b7ee9/packages/%40aws-cdk/aws-ec2/lib/vpc.ts#L524-L529

test/unit/vpc.test.ts
```diff
@@ -0,0 +1,75 @@
+import { get } from "lodash";
+import { merge } from "lodash";
+import { baseConfig, pluginConfigExt, runServerless } from "../utils/runServerless";
+import ServerlessError from "../../src/utils/error";
+import {AwsCfInstruction} from "@serverless/typescript";
+
+describe("vpc", () => {
+    it("should have subnets and security groups", async () => {
```
mnapoli, line 8: If I understand correctly what this test creates, it could be improved to:

```suggestion
    it("should put Lambda functions in the VPC", async () => {
```

test/unit/vpc.test.ts
```diff
@@ -0,0 +1,75 @@
...
+                        },
+                    },
+                }),
+                command: "package",
+            });
+        } catch (error) {
+            expect(error).toBeInstanceOf(ServerlessError);
+            expect(error).toHaveProperty("code", "LIFT_ONLY_ONE_VPC");
+        }
+    });
```
mnapoli, line 74: Awesome testing the error cases. I would also add a `should create all required resources` kind-of test, like for the other construct: i.e. test that `serverless deploy` actually deploys a VPC with all the related resources in the CloudFormation template.

mnapoli (commented):
That looks awesome, thank you!

I've added minor comments and then for me we'll be good to merge!

src/classes/AwsConstruct.ts
t-richard: Wasn't aware of that, I applied those changes my self. I will try to rebase.

src/constructs/VPC.ts
t-richard: They indeed seem similar, so let's go with the more concise `vpc.privateSubnets` syntax.

test/unit/vpc.test.ts
t-richard: Isn't it testing the CDK more than testing Lift ? I'll see what I can come up with.

src/classes/AwsProvider.ts
t-richard: That's more of an idea for later. 

Security groups are not related to a specific subnet/vpc so one may want to provide a custom security group that is shared accross projets/environments and we might want to merge our newly created one with those provided by the user.

src/constructs/VPC.ts
```diff
@@ -0,0 +1,52 @@
...
+
+        const lambdaSecurityGroup = new SecurityGroup(this, "AppSecurityGroup", {
```
t-richard, line 35: I think we need to add an egress here to allow outbound trafic on http/https

src/classes/AwsConstruct.ts
t-richard: I don't see those on master. It should be here for example https://github.com/getlift/lift/blob/master/src/classes/AwsConstruct.ts#L32

Let me know if I'm missing something.

src/constructs/VPC.ts
```diff
@@ -0,0 +1,50 @@
...
+    private readonly vpc: Vpc;
+
+    constructor(scope: CdkConstruct, id: string, configuration: Configuration, private provider: AwsProvider) {
+        super(scope, id);
+
+        this.vpc = new Vpc(this, "VPC", {
+            maxAzs: 2,
+        });
+
+        const privateSubnets = this.vpc.privateSubnets;
```
t-richard, line 31: Also maybe we would like to create isolated subnets to put services into (like RDS, ElastiCache, etc), instead of putting them in the private subnets.

src/classes/AwsConstruct.ts
mnapoli: 🤦  my bad those were in the `db` branch. All good!

src/classes/AwsProvider.ts
mnapoli: 👍  ok makes sense

src/constructs/VPC.ts
mnapoli: Yes, I would even say allow all outbound traffic.

test/unit/vpc.test.ts
mnapoli: 👍  have a look at the other constructs

The idea is to test what is being deployed, no matter if it's using the default values of CDK constructs or if we customize it. Lift provides the outcome, and unlike CloudFormation CDK constructs could evolve.

src/constructs/VPC.ts
mnapoli: Let's put that aside for now, putting these services in the private subnets should be a good fit for most use cases.

@deleugpn mentioned that isolated subnets could be useful for security compliance, but these are advanced use cases.

src/constructs/VPC.ts
t-richard: Done in the latest commit

test/unit/vpc.test.ts
t-richard: Done in the latest commit

t-richard:
Added tests to ensure resources are created and added the security group egress.

Also removed the calls to `provider.getCloudformationReference()` to be ready for https://github.com/getlift/lift/pull/71

I think the only missing thing is docs now. I'll work on that tonight and we should be good to go.

mnapoli:
As discussed in Slack, merged as a "shadow" construct: it is not documented. It is experimental for now, and will be iterated as we identify use cases around it (e.g. RDS, etc.).

PR #67, opened 2021-07-04
New `queue` commands: logs and "send message"

mnapoli:
This PR introduces 2 new `queue` commands to facilitate user onboarding:

- `serverless <construct-name>:logs`

This command displays the logs of the Lambda "worker" function.

It is an alias to `serverless logs --function <construct-name>Worker` and supports the same options, for example `--tail` to tail logs live.

- `serverless <construct-name>:send`

Send a message into the SQS queue.

This command can be useful while developing to push sample messages into the queue.

When the command runs, it will prompt for the body of the SQS message. It is also possible to provide the body via the `--body="message body here"` option.


fredericbarthelet (approved):

mnapoli:
image

The body can also be provided interactively:

Screen 20210705 zgUU91f8

PR #69, opened 2021-07-05
Single table construct

fredericbarthelet:
This PR implements one of the use case identified in https://github.com/getlift/lift/discussions/54, the Single Table Design constructs.

# Database - DynamoDB Single Table

The `database/dynamodb-single-table` construct deploys a single DynamoDB table with pre-configured indexes following [Single Table Design](https://www.alexdebrie.com/posts/dynamodb-single-table/) principles.

## Quick start

```yaml
service: my-app
provider:
  name: aws

constructs:
    myTable:
        type: database/dynamodb-single-table

plugins:
    - serverless-lift
```

On `serverless deploy`, a preconfigured DynamoDB table will be created.

## How it works

The `database/dynamodb-single-table` construct creates and configures the table for production:

- a composite primary index with generic attributes names - `PK` for the partition key and `SK` for the sort key
- 5 local secondary indexes, the maximum allowed, with generic names - `LSI-1` to `LSI-5` for the index names and `LSI-1-SK` to `LSI-5-SK` for the sort keys. Those indexes have no impact on pricing as long as their indexes are not populated with data. Modifying a table local secondary indexes configuration requires table re-creation, thus enabling all of them by default at table creation.
- a configurable amount of up to 20 glocal secondary indexes with generic names - `GSI-1` to `GSI-20` for the index names, `GSI-1-PK` to `GSI-20-PK` for the partition keys and `GSI-1-SK` to `GSI-20-SK` for the sort keys
- all indexes attributes have string data type, ideal for composite attribue - i.e. `value1#value2`
- a DynamoDB stream publishing new and old values at each write operation on the table

## Variables

All database constructs expose the following variables:

- `tableName`: the name of the deployed DynamoDB table
- `tableStreamArn`: the ARN of the stream of the deployed DynamoDB table

This can be used to inject the tableName to a Lambda functions using the SDK to read or write data from the table, for example:

```yaml
constructs:
    myTable:
        type: database/dynamodb-single-table

functions:
    myFunction:
        handler: src/index.handler
        environment:
            TABLE_NAME: ${construct:myTable.tableName}
```

_How it works: the `${construct:myTable.tableName}` variable will automatically be replaced with a CloudFormation reference to the DynamoDB table._

## Permissions

By default, all the Lambda functions deployed in the same `serverless.yml` file **will be allowed to read/write into the table**.

In the example below, there are no IAM permissions to set up: `myFunction` will be allowed to read and write into the `myTable` table.

```yaml
constructs:
    myTable:
        type: database/dynamodb-single-table

functions:
    myFunction:
        handler: src/index.handler
        environment:
            TABLE_NAME: ${construct:myTable.tableName}
```

### Global secondary indexes

Global secondary indexes have a direct impact on the cost of a DynamoDB table. There is no GSI configured by default on the database construct.

You can specify the amount of GSI you'd like to enable on a DynamoDB table using the `gsiCount` property.

```yaml
constructs:
    myTable:
        # ...
        gsiCount: 3
```

The first time you deploy your construct using `serverless deploy`, you can specify any amount of GSI between `1` and `20`. On subsequent deploys, any modification made to an already deployed construct cannot add or remove more than 1 GSI at a time. If you need 2 additional GSI after initial deployment of the exemple above, you must first update the `gsiCount` to `4`, deploy, and then finally update it to the final desired quantity of `5`.

### DIFF
--- a/docs/database-dynamodb-single-table.md
+++ b/docs/database-dynamodb-single-table.md
@@ -24,7 +24,6 @@ On `serverless deploy`, a preconfigured DynamoDB table will be created.
 The `database/dynamodb-single-table` construct creates and configures the table for production following [Single Table Design](https://www.alexdebrie.com/posts/dynamodb-single-table/) principles:
 
 - a composite primary index with generic attributes names - `PK` for the partition key and `SK` for the sort key
-- 5 local secondary indexes, the maximum allowed, with generic names - `LSI-1` to `LSI-5` for the index names and `LSI-1-SK` to `LSI-5-SK` for the sort keys. Those indexes have no impact on pricing as long as their indexes are not populated with data. Modifying a table local secondary indexes configuration requires table re-creation, thus enabling all of them by default at table creation.
 - a configurable amount of up to 20 glocal secondary indexes with generic names - `GSI-1` to `GSI-20` for the index names, `GSI-1-PK` to `GSI-20-PK` for the partition keys and `GSI-1-SK` to `GSI-20-SK` for the sort keys
 - all indexes attributes have string data type, ideal for composite attribue - i.e. `value1#value2`
 - a DynamoDB stream publishing new and old values at each write operation on the table
@@ -86,6 +85,23 @@ constructs:
 ```
 
 The first time you deploy your construct using `serverless deploy`, you can specify any amount of GSI between `1` and `20`. On subsequent deploys, any modification made to an already deployed construct cannot add or remove more than 1 GSI at a time. If you need 2 additional GSI after initial deployment of the exemple above, you must first update the `gsiCount` to `4`, deploy, and then finally update it to the final desired quantity of `5`.
+
+### Local secondary indexes
+
+Each DynamoDB table can includes up to 5 local secondary indexes. You can deploy a table with those 5 indexes using the `localSecondaryIndexes` property.
+
+> :warning: LSIs introduce a [limitation on partition size of a table](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/LSI.html#LSI.ItemCollections.SizeLimit). Due to this limitation, `localSecondaryIndexes` is set to false and this construct will not provision any LSI on the table by default.
+
+Setting `localSecondaryIndexes` to true will provision 5 LSIs with generic names - `LSI-1` to `LSI-5` for the index names and `LSI-1-SK` to `LSI-5-SK` for the sort keys. Those indexes have no impact on pricing as long as their sort keys are not populated with data.
+
+```yaml
+constructs:
+    myTable:
+        # ...
+        localSecondaryIndexes: true
+```
+
+> :warning: Modifying a table local secondary indexes configuration requires table re-creation. If you modify this setting after the table has been populated with data, you'll need to transfer all data from old table to the new one. You however won't loose any data as all tables are configured to be left as is when removed from a CloudFormation template.
 
 ### More options
 
--- a/docs/database-dynamodb-single-table.md
+++ b/docs/database-dynamodb-single-table.md
@@ -21,7 +21,7 @@ On `serverless deploy`, a preconfigured DynamoDB table will be created.
 
 ## How it works
 
-The `database/dynamodb-single-table` construct creates and configures the table for production:
+The `database/dynamodb-single-table` construct creates and configures the table for production following [Single Table Design](https://www.alexdebrie.com/posts/dynamodb-single-table/) principles:
 
 - a composite primary index with generic attributes names - `PK` for the partition key and `SK` for the sort key
 - 5 local secondary indexes, the maximum allowed, with generic names - `LSI-1` to `LSI-5` for the index names and `LSI-1-SK` to `LSI-5-SK` for the sort keys. Those indexes have no impact on pricing as long as their indexes are not populated with data. Modifying a table local secondary indexes configuration requires table re-creation, thus enabling all of them by default at table creation.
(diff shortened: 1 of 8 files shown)


docs/database-dynamodb-single-table.md
```diff
@@ -0,0 +1,92 @@
...
+
+plugins:
+    - serverless-lift
+```
+
+On `serverless deploy`, a preconfigured DynamoDB table will be created.
+
+## How it works
+
+The `database/dynamodb-single-table` construct creates and configures the table for production:
```
mnapoli, line 24: Even though it's repeating the introduction paragraph, this piece of information might be very important to help new users understand why everything is so locked down "in a weird way" (at least until you've learned what single-table design is). So that might be worth repeating:

```suggestion
The `database/dynamodb-single-table` construct creates and configures the table for production following [Single Table Design](https://www.alexdebrie.com/posts/dynamodb-single-table/) principles:
```

docs/database-dynamodb-single-table.md
```diff
@@ -0,0 +1,92 @@
...
+
+- a composite primary index with generic attributes names - `PK` for the partition key and `SK` for the sort key
+- 5 local secondary indexes, the maximum allowed, with generic names - `LSI-1` to `LSI-5` for the index names and `LSI-1-SK` to `LSI-5-SK` for the sort keys. Those indexes have no impact on pricing as long as their indexes are not populated with data. Modifying a table local secondary indexes configuration requires table re-creation, thus enabling all of them by default at table creation.
+- a configurable amount of up to 20 glocal secondary indexes with generic names - `GSI-1` to `GSI-20` for the index names, `GSI-1-PK` to `GSI-20-PK` for the partition keys and `GSI-1-SK` to `GSI-20-SK` for the sort keys
+- all indexes attributes have string data type, ideal for composite attribue - i.e. `value1#value2`
+- a DynamoDB stream publishing new and old values at each write operation on the table
```
mnapoli, line 30: BTW do any of these settings incur additional costs? (for example the streams?)

docs/database-dynamodb-single-table.md
```diff
@@ -0,0 +1,92 @@
...
+
+Global secondary indexes have a direct impact on the cost of a DynamoDB table. There is no GSI configured by default on the database construct.
+
+You can specify the amount of GSI you'd like to enable on a DynamoDB table using the `gsiQuantity` property.
+
+```yaml
+constructs:
+    myTable:
+        # ...
+        gsiQuantity: 3
```
mnapoli, line 85: Nothing major but I'm wondering if `gsiCount` would work better? If any native English speaker can chime in here that would be great.

I'd say "count" is more widespread in developer language, but maybe "quantity" is used in DynamoDB docs/settings already?

Anyway, I'm fine either way, I'm just opening the topic just in case.

mnapoli (commented):
This is awesome!

docs/database-dynamodb-single-table.md
fredericbarthelet: AFAIK, streams does not incur additional costs

docs/database-dynamodb-single-table.md
fredericbarthelet: I prefer `gsiCount` as well, updated it :)

alexdebrie:
Hey @fredericbarthelet, this is cool! A few thoughts as I look it over:

1. I wouldn't add the LSIs by default. They add a [weird limitation on item collection max size](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/LSI.html#LSI.ItemCollections.SizeLimit), which could be very bad in certain cases. I almost never use LSIs.

2. I assume the billing mode is `PAY_PER_REQUEST`? Is this configurable?

3. Do the permissions extend to all secondary indexes as well? Do they allow all read + write operations? Some people could recoil a bit at this, but maybe it's the philosophy of Lift to do it this way?

fredericbarthelet:
Thanks for your feedback @alexdebrie :)

I updated the PR to reflect the following:
-  no LSI by default, it is now an opt-in parameter
- the billing mode is PAY_PER_REQUEST, it is not configurable at the time being
- read/write permissions are enabled by default on primary and secondary indexes for all lambdas provisioned in the same serverless.yml file. This behavior is the default but will be opt-out on user request (we're planning on adding a general setting to not provision lambda's permission).
- secondary keys are not configurable. I was wondering if duplicating data across multiple attribute when there is an overlap in indexes is not a better strategy. For exemple, you may have access patterns with overlapping indexes for a specific type of entities within your table. If you enforce GSI-1-PK as being the table primary SK, you actually enforce projection of all rows of the table to this new secondary index, even for entities not benefiting from this pattern. This incurs additional cost, that may be higher than actually duplicating this specific attribute for table benefitting from this pattern.

*Solution with attribute duplication*
| PK         	| SK       	| GSI-1-PK 	| GSI-1-SK   	| Amount 	|
|------------	|----------	|----------	|------------	|--------	|
| INV#123456 	| BILL#112 	| BILL#112 	| 2020-01-03 	| 12     	|
| INV#123456 	| BILL#113 	| BILL#113 	| 2020-01-08 	| 22     	|
| USER#123   	| John Doe 	|          	|            	|        	|

Benefits:
- Inserting USR#123 does not incur cost for replication on GSI-1
- Invoices have a duplicated attribute: SK and GSI-1-PK. This incur additional costs only if this duplication actually brings the total amount written over the next KB

*Solution with overlapping indexes*
| PK         	| SK (GSI-1-PK) 	| GSI-1-SK   	| Amount 	|
|------------	|---------------	|------------	|--------	|
| INV#123456 	| BILL#112      	| 2020-01-03 	| 12     	|
| INV#123456 	| BILL#113      	| 2020-01-08 	| 22     	|
| USER#123   	| John Doe      	|            	|        	|

Benefits:
- Duplication does not need to be programmatically handled

WDYT ?

fredericbarthelet:
Merging this as is for now, @alexdebrie we can continue the conversation in https://github.com/getlift/lift/issues/76

richard457:
Hi, I used to reference table like this ```EVENT:
    handler: EVENT.handler
    tracing: "false"
    events:
      - stream:
          type: "dynamodb"
          batchSize: 1
          startingPosition: LATEST
          arn:
            Fn::GetAtt:
              - Table
              - StreamArn
          #       # https://blog.devgenius.io/how-to-filter-serverless-events-from-dynamodb-streams-c5f7c7250674
          filterPatterns:
            - eventName: [INSERT]``` but now I want to use table generated by construct how can I do that?

mnapoli:
@richard457 let's keep the discussion in #282

PR #71, opened 2021-07-07
Simplify variables by using CDK values directly

mnapoli:
On `master`, construct variables are defined as CloudFormation references (or "GetAtt"). These references are generated from CDK tokens, which come from CDK object properties.

Since we already resolve all CDK tokens that appear in `serverless.yml`, we can skip some work and use the CDK variables directly.

Why:

- CDK constructs are simpler to write, and simpler to reason about
- importing existing CDK constructs as custom constructs (if we decide to merge that feature) is simplified too
- AWS variables (CF references) and non-AWS variables (i.e. if we ever support multiprovider) are unified, or at least that's the idea (we'll see in practice when we get to it)

### DIFF
--- a/src/classes/Construct.ts
+++ b/src/classes/Construct.ts
@@ -6,12 +6,15 @@ import { CliOptions } from "../types/serverless";
  * Defines which methods a Lift construct must expose.
  */
 export interface ConstructInterface {
-    outputs(): Record<string, () => Promise<string | undefined>>;
+    /**
+     * Values shown in the CLI output.
+     */
+    outputs?(): Record<string, () => Promise<string | undefined>>;
 
     /**
-     * CloudFormation references
+     * serverless.yml variables
      */
-    references(): Record<string, Record<string, unknown>>;
+    variables?(): Record<string, unknown>;
 
     /**
      * Post-CloudFormation deployment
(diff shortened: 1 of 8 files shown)


src/classes/Construct.ts
```diff
@@ -6,12 +6,15 @@ import { CliOptions } from "../types/serverless";
  * Defines which methods a Lift construct must expose.
  */
 export interface ConstructInterface {
-    outputs(): Record<string, () => Promise<string | undefined>>;
+    /**
+     * Values shown in the CLI output.
+     */
+    outputs?(): Record<string, () => Promise<string | undefined>>;
```
mnapoli, line 12: I've taken the opportunity to mark all functions as optional. Some constructs don't expose outputs or variables (just like commands), so that simplifies writing constructs.

src/classes/Construct.ts
```diff
@@ -6,12 +6,15 @@ import { CliOptions } from "../types/serverless";
...
 
     /**
-     * CloudFormation references
+     * serverless.yml variables
      */
-    references(): Record<string, Record<string, unknown>>;
+    variables?(): Record<string, unknown>;
```
mnapoli, line 17: Renamed to `variables` because now that is not specifically a cloudformation reference.

t-richard (approved):

PR #74, opened 2021-07-12
Stripe provider

fredericbarthelet:
First implementation of https://github.com/getlift/lift/discussions/61


src/plugin.ts
```diff
@@ -53,6 +77,7 @@ class LiftPlugin {
 
     constructor(serverless: Serverless, cliOptions: Record<string, string>) {
         this.serverless = serverless;
+        Object.assign(this.serverless, { getProviderById: this.getProviderById.bind(this) });
```
mnapoli, line 80: `getProviderById` -> `getLiftProviderById`?

To avoid any surprise/conflict, because Lift providers are not the same as the existing provider system.

Also it seems like it's added for tests only. It might be worth documenting this, e.g.

```suggestion
        // @internal This method is exposed for Lift tests only, it is not a public API
        Object.assign(this.serverless, { getProviderById: this.getProviderById.bind(this) });
```

src/plugin.ts
```diff
@@ -1,22 +1,44 @@
...
     VariableResolver,
 } from "./types/serverless";
-import { AwsProvider, ConstructInterface } from "./classes";
 import { log } from "./utils/logger";
-import { StaticConstructInterface } from "./classes/Construct";
 import ServerlessError from "./utils/error";
 
+const PROVIDER_ID_PATTERN = "^[a-zA-Z0-9-_]+$";
+// This is enable all existing constructs defined prior intoduction of "providers" property to work
+const DEFAULT_PROVIDER = "defaultAwsProvider";
```
mnapoli, line 23: Shouldn't the default provider be called `aws`?

src/providers/AwsProvider.ts
```diff
@@ -1,20 +1,26 @@
...
-import { Webhook } from "../constructs/Webhook";
-import { StaticWebsite } from "../constructs/StaticWebsite";
-import { DatabaseDynamoDBSingleTable } from "../constructs/DatabaseDynamoDBSingleTable";
-
-export class AwsProvider {
+
+const AWS_DEFINITION = {
+    type: "object",
+    properties: {
+        profile: { type: "string" },
```
mnapoli, line 16: Isn't the `aws` provider not configurable in the `providers` section?

If yes, that line should probably be removed right?

src/providers/ProviderInterface.ts
```diff
@@ -0,0 +1,5 @@
+import type { ConstructInterface } from "@lift/constructs";
+
+export interface ProviderInterface {
+    create(type: string, id: string): ConstructInterface;
```
mnapoli, line 4: This seems to conflict with the static `create()` method. Maybe we should name it `createConstruct()`?

src/providers/StripeProvider.ts
```diff
@@ -0,0 +1,137 @@
...
+    static getAllConstructClasses(): StaticConstructInterface[] {
+        return Object.values(this.constructClasses);
+    }
+
+    static create(serverless: Serverless, id: string, { profile }: Configuration): StripeProvider {
+        return new this(serverless, id, profile);
+    }
+
+    private config: { apiKey: string; accountId?: string };
+    public provider: Stripe;
```
mnapoli, line 63:
```suggestion
    public sdk: Stripe;
```

to avoid any confusion with the "provider" concept we already have? (i.e. here we have a provider in a provider, that's confusing)

src/providers/StripeProvider.ts
```diff
@@ -0,0 +1,137 @@
...
+    create(type: string, id: string): ConstructInterface {
+        const Construct = StripeProvider.getConstructClass(type);
+        if (Construct === undefined) {
+            throw new ServerlessError(
+                `The construct '${id}' has an unknown type '${type}'\n` +
+                    "Find all construct types available here: https://github.com/getlift/lift#constructs",
+                "LIFT_UNKNOWN_CONSTRUCT_TYPE"
+            );
+        }
+        const configuration = get(this.serverless.configurationInput.constructs, id, {});
```
mnapoli, line 78: Would it make sense for the configuration of the construct to be passed to the method as a parameter, rather than have this method fetch the config from the `constructs` section on the `serverless` object?

I.e. the signature of the method could be something like: `create(type: string, id: string, configuration: Record<...>): ConstructInterface`

mnapoli (approved):
🎉

PR #75, opened 2021-07-13
Allow disabling automatic permissions

mnapoli:
```yaml
# serverless.yml

lift:
    automaticPermissions: false
```

This new option allows to opt-out from automatic permissions set up by Lift constructs.

### DIFF
--- a/docs/queue.md
+++ b/docs/queue.md
@@ -146,6 +146,8 @@ functions:
             QUEUE_URL: ${construct:my-queue.queueUrl}
 ```
 
+Automatic permissions can be disabled: [read more about IAM permissions](permissions.md).
+
 ## Commands
 
 The following commands are available on `queue` constructs:
--- a/docs/permissions.md
+++ b/docs/permissions.md
@@ -1,0 +1,55 @@
+# IAM Permissions
+
+Lift constructs are designed to be functional out of the box.
+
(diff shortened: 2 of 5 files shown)


fredericbarthelet (approved):
Nice one :)

docs/queue.md
```diff
@@ -146,6 +146,8 @@ functions:
             QUEUE_URL: ${construct:my-queue.queueUrl}
 ```
 
+Automatic permissions can be disabled: [read more about IAM permissions](permissions.md).
```
t-richard, line 149:
```suggestion
Automatic permissions can be disabled: [read more about IAM permissions](permissions.md#disabling-automatic-permissions).
```

t-richard (commented):
Nice :+1:

docs/queue.md
mnapoli: The link is intentional ("read more about IAM permissions"): the separate page recaps how permissions work in Lift in general.

I'd rather have users read the whole page to understand how it works. At the moment the documentation on IAM permissions is a bit vague (i.e. "it works magically"), so I think that separate page should do 2 jobs:

- explain how permissions work (to clear up confusion)
- show how to disable

t-richard: I get your point, it makes sense ! :+1:

t-richard (approved):

PR #87, opened 2021-08-24
Stripe construct

fredericbarthelet:


src/providers/StripeProvider.ts
```diff
@@ -61,9 +62,22 @@ export class StripeProvider implements ProviderInterface {
 
     private config: { apiKey: string; accountId?: string };
     public sdk: Stripe;
+    private state: Record<string, unknown>;
```
mnapoli, line 65: Do we need state management here? I'm wondering if this is a bit early to add.

PR #88, opened 2021-08-24
Support FIFO Queues

AndrewBarba:
- Adds support for FIFO queues

### DIFF
--- a/src/constructs/aws/Queue.ts
+++ b/src/constructs/aws/Queue.ts
@@ -222,7 +222,7 @@ export class Queue extends AwsConstruct {
                     arn: this.queue.queueArn,
                     batchSize: batchSize,
                     // TODO add setting
-                    maximumBatchingWindow: 60,
+                    maximumBatchingWindow: this.queue.fifo ? undefined : 60,
                 },
             },
         ];
--- a/src/constructs/aws/Queue.ts
+++ b/src/constructs/aws/Queue.ts
@@ -118,9 +118,10 @@ export class Queue extends AwsConstruct {
         const baseName = `${this.provider.stackName}-${id}`;
 
         const dlq = new CdkQueue(this, "Dlq", {
-            queueName: `${baseName}-dlq`,
+            queueName: configuration.fifo === true ? `${baseName}-dlq.fifo` : `${baseName}-dlq`,
             // 14 days is the maximum, we want to keep these messages for as long as possible
             retentionPeriod: Duration.days(14),
+            fifo: configuration.fifo,
         });
 
         this.queue = new CdkQueue(this, "Queue", {
--- a/src/constructs/aws/Queue.ts
+++ b/src/constructs/aws/Queue.ts
@@ -126,6 +126,7 @@ export class Queue extends AwsConstruct {
             // 14 days is the maximum, we want to keep these messages for as long as possible
             retentionPeriod: Duration.days(14),
             fifo: configuration.fifo,
+            contentBasedDeduplication: configuration.fifo,
         });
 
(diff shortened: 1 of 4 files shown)


src/constructs/aws/Queue.ts
```diff
@@ -114,21 +115,24 @@ export class Queue extends AwsConstruct {
 
         const maxRetries = configuration.maxRetries ?? 3;
 
+        const baseName = `${this.provider.stackName}-${id}`;
+
         const dlq = new CdkQueue(this, "Dlq", {
-            queueName: `${this.provider.stackName}-${id}-dlq`,
+            queueName: `${baseName}-dlq`,
             // 14 days is the maximum, we want to keep these messages for as long as possible
             retentionPeriod: Duration.days(14),
```
mnapoli, line 123: We probably need the DLQ to be configured as FIFO as well, right?

AndrewBarba: I wasn't entirely sure what the best practice is there, but thinking about it that does make sense to me. I'm going to update it

AndrewBarba: Ah never mind:

> The dead-letter queue of a FIFO queue must also be a FIFO queue. Similarly, the dead-letter queue of a standard queue must also be a standard queue.

mnapoli:
Thanks!

I'm wondering what the impact would be on:

- `serverless myqueue:send` to send a message
    Users would need to be able to pass the message group ID right?
- `serverless myqueue:failed`
    If the DLQ is FIFO, then that command will not be able to list all messages? It will only be able to list the first one for each message group I expect?

AndrewBarba:
@mnapoli 

1. Correct - seems useful to be able to send that
2. I dont think thats the case as described here: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues-understanding-logic.html

> It is possible to receive up to 10 messages in a single call using the MaxNumberOfMessages request parameter of the ReceiveMessage action. These messages retain their FIFO order and can have the same message group ID. Thus, if there are fewer than 10 messages available with the same message group ID, you might receive messages from another message group ID, in the same batch of 10 messages, but still in FIFO order.

AndrewBarba:
@mnapoli I just added cli support for `MessageGroupId` and added docs to readme

AndrewBarba:
@mnapoli @fredericbarthelet  I enabled content deduplication and updated tests so all pass.

Let me know if you need anything else before merging

src/constructs/aws/Queue.ts
```diff
@@ -114,21 +119,27 @@ export class Queue extends AwsConstruct {
...
 
+        const baseName = `${this.provider.stackName}-${id}`;
+
         const dlq = new CdkQueue(this, "Dlq", {
-            queueName: `${this.provider.stackName}-${id}-dlq`,
+            queueName: configuration.fifo === true ? `${baseName}-dlq.fifo` : `${baseName}-dlq`,
             // 14 days is the maximum, we want to keep these messages for as long as possible
             retentionPeriod: Duration.days(14),
+            fifo: configuration.fifo,
+            contentBasedDeduplication: configuration.fifo,
```
mnapoli, line 129: I don't think the DLQ should deduplicate messages based on content automatically: this is a queue meant to store failed messages over a longer period of time, so we don't want similar messages to be deduplicated.

I'll remove that in the merge process, let me know if you think that's wrong.

mnapoli:
Thank you for the contribution!

Just merged it to master manually with more tests and docs, I'll close the PR manually.

PR #90, opened 2021-08-29
[RFC] MySQL database construct

mnapoli:
This PR builds on the VPC (#65) to add a MySQL database instance.

Example:

```yaml
constructs:
    database:
        type: database/mysql/dev
        password: qwertyqwerty
```

Some lessons while working on this (up for discussion of course):

- I've turned the VPC from a construct to a "special" object on the `AwsProvider` class
  I realize this is opposite to what we previously agreed to in https://github.com/getlift/lift/discussions/66#discussioncomment-948684 I'm not 100% sure about that yet, but I've detailed in the code the reasons why: https://github.com/getlift/lift/blob/8e60ad80b042f9c9e7fdf09950d2bbab97377daf/src/providers/AwsProvider.ts#L98-L112
  This isn't perfect, but at least that works and provides a good user experience.
- The construct is very very specific: RDS, no Aurora, no Aurora serverless, no Postgre, no cluster… It's a MySQL instance only.
  Initially, that construct was supposed to be called `database`. But it means so many different things. With the addition of `database/dynamodb-single-table`, it made sense to have something like `database/rds`. But then I realized: that doesn't really mean anything for developers. As a developer, I want "a MySQL database". Focusing on a single engine simplified the code a lot, and I think it will simplify the config as well. The downside is possibly having multiple constructs, but at least those constructs are very specific and close to the use case.
  Additionally, I think dev databases and prod databases have completely different configs. It's not a matter of configuring a few options, but it's entirely different. That's why I think it wouldn't be crazy to have `database/mysql/dev` and `database/mysql` (for prod). At the moment the PR only implements the dev version, because it's still very early.

As you can probably tell, that PR is extremely draft. But I know @t-richard was interested at some point, and why not get some discussions started.


t-richard:
Thanks for starting this :raised_hands: :partying_face: 

> I've turned the VPC from a construct to a "special" object on the AwsProvider class

This is indeed the only way. I started to work on a similar construct and stumbled on the exact same problem. Even if it feels less flexible, I agree it makes more sense from a DX and maintenance POV. 100% with you on that, sorry for suggesting it should be a construct in the first place :joy: 

> The construct is very very specific: RDS, no Aurora, no Aurora serverless, no Postgre, no cluster…

I agree it makes more sense to have something separated for Aurora clusters and Aurora serverless (and probabably separated between v1 and v2 too) because they are very different to configure and are targeted to different use-cases. But making it MySQL only seems weird to me because PostgreSQL and MariaDB are so close in terms of RDS configuration and have been getting a lot of traction in the latest years. 

For example, Symfony made Postgres the default choice for most use-cases and added a lot of Postgres-specific integrations for queues, locks, etc. Platforms like Heroku have a built-in Heroku Postgres add-on but the MySQL ones are maintained by third-parties. Postgres usage I huge in communities like Ruby

If you take the latest [Stackoverflow survey](https://insights.stackoverflow.com/survey/2021#section-most-popular-technologies-databases), MySQL and PostgreSQL usage are relatively close (50% vs 40%), [MySQL is more dreaded by its users than Postgres](https://insights.stackoverflow.com/survey/2021#section-most-loved-dreaded-and-wanted-databases) and [2x more people using MySQL want to use Postgres than the other way around](https://insights.stackoverflow.com/survey/2021#section-most-loved-dreaded-and-wanted-databases).

To be clear, I'm not saying MySQL is bad or Postgres is way better. Just the fact that thinking MySQL is the defacto choice is not true anymore. 

I'd rather have a generic `database/rdbms` or `database/sql` or something similar with restrained config options like `engine` and `engineVersion`. 

Or maybe we could come up with a generic base construct that would be "abstract" and have engine specific constructs that would just provide the values to use to the base one. Not sure how Typescript would work with this though.

Going this way, I fear it will just multiply and will be hard to maintain in the long run. 
Imagine being a year from now, with constructs for RDS, Aurora, Aurora Serverless v1, Aurora Serverless v2 for each of the 2/3 (or even 5) engines available, that seems like a mess to maintain to me.

> That's why I think it wouldn't be crazy to have database/mysql/dev and database/mysql

How would that work to deploy a different type to a different stage ? Like `serverless deploy --stage=prod` uses the prod one and the rest is using dev ?

I'd be happy to discuss it. Let me know what you think :slightly_smiling_face:

mnapoli:
> This is indeed the only way. I started to work on a similar construct and stumbled on the exact same problem. Even if it feels less flexible, I agree it makes more sense from a DX and maintenance POV. 100% with you on that, sorry for suggesting it should be a construct in the first place 😂

Ah no :) We've learned and we are now much more confident with that choice, this is a win!

Regarding MySQL vs PostgreSQL first: TIL that Postgre was so popular. In itself, I have no problem focusing on that engine first, or even releasing MySQL + Postgre in the first version.

Now on the second aspect: "should it be a single component or 2". TBH I'm not 100% sure, but as a developer, using a `database/mysql` construct is much more obvious than `database/rds` or `database/sql`. Maybe I'm taking it a bit too far though.

In terms of maintenance, I'm not sure it would be that much of a problem because we could have 1 parent class/share code.

What actually tipped me over is that logging configuration was different (I had to add an "if mysql then ... else ..."). Now I explored [the full list of settings](https://docs.aws.amazon.com/fr_fr/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html) and it turns out there are very few differences…

So let's try to merge mysql + mariadb + postgresql into a single component. I do like the `sql` name, previously I had `rds` in mind (or `rdbms`) but "sql" is much better.

> How would that work to deploy a different type to a different stage ? Like serverless deploy --stage=prod uses the prod one and the rest is using dev ?

Yeah, this one is a tough one. I have options in mind but nothing that's really perfect. The goal is allowing to deploy dev/staging environments without the delay and extra costs of a production setup (would also apply to websites to ignore cloudfront for example), all ideas are welcome :)

docs/database-sql.md
```diff
@@ -0,0 +1,209 @@
...
+
+The secret can then be retrieved at runtime by the Lambda function:
+
+```js
+const AWS = require('aws-sdk');
+const secretsmanager = new AWS.SecretsManager();
+const secret = await client.getSecretValue({
+    SecretId: process.env.DB_PASSWORD_SECRET,
+}).promise();
+const dbPassword = secret.SecretString;
```
mnapoli, line 204: TODO actually test that (because I'm 100% sure it's not finished)

docs/database-sql.md
```diff
@@ -0,0 +1,209 @@
...
+```
```
mnapoli, line 205: TODO: document `storageSize`

src/constructs/aws/Vpc.ts
```diff
@@ -20,12 +20,12 @@ export class Vpc extends CdkVpc implements ConstructInterface {
...
+        return new this(provider.stack, id, configuration);
     }
 
-    private readonly appSecurityGroup: SecurityGroup;
+    public readonly appSecurityGroup: SecurityGroup;
 
-    constructor(scope: CdkConstruct, id: string, configuration: Configuration, private provider: AwsProvider) {
+    constructor(scope: CdkConstruct, id: string, private configuration: Configuration) {
         super(scope, id, {
             maxAzs: 2,
```
mnapoli, line 30: TODO: document the VPC

For example by default it deploys 2 NAT Gateway

docs/database-sql.md
```diff
@@ -0,0 +1,209 @@
...
+        type: database/sql
+
+functions:
+    myfunction:
+        # ...
+        environment:
+            DB_HOST: ${construct:db.host}
+```
+
+_How it works: the `${construct:db.queueUrl}` variable will automatically be replaced with a CloudFormation reference to the database._
```
t-richard, line 109:
```suggestion
_How it works: the `${construct:db. 
host}` variable will automatically be replaced with a CloudFormation reference to the database._
```

src/constructs/aws/DatabaseSql.ts
```diff
@@ -0,0 +1,161 @@
...
+            vpcSubnets: {
+                subnetType: SubnetType.PRIVATE,
+            },
+            // We go with 20 (the minimum) instead of 100 by default, because it is much cheaper
+            allocatedStorage: configuration.storageSize ?? 20,
+            // 7 days backups instead of the default 1
+            backupRetention: Duration.days(7),
+            cloudwatchLogsExports: this.logsToEnable(),
+        });
+
```
t-richard, line 86: Maybe we need to add a security group to allow inbound from the lambda security group?

src/constructs/aws/DatabaseSql.ts
```diff
@@ -0,0 +1,161 @@
...
+    outputs(): Record<string, () => Promise<string | undefined>> {
+        return {
+            host: () => this.provider.getStackOutput(this.dbHostOutput),
+        };
+    }
+
+    variables(): Record<string, unknown> {
+        return {
+            host: this.dbInstance.instanceEndpoint.hostname,
```
t-richard, line 118: It would be nice to construct a connection string because lots of frameworks/libraries encourage to use them by default (something like `mysql://user:pass@host:port/dbName`)

I searched a bit and the format seems consistent across libraries. The only exception is Java JDBC which is mostly the same but prefixed with `jdbc:`

t-richard:
This looks promising 😁

I just fear the version might cause troubles.

I think I had cases with cloudformation where
1. I created a db with version x
2. Aws updated to version y during a maintenance window
3. I updated my db in cloudformation (like changing storage size) but still used version x
4. It failed

I'll have to test again but it may be a problem if the user has no control over it.

I think we could also add encryption or multi AZ but it could be done in a separate PR.

src/constructs/aws/DatabaseSql.ts
```diff
@@ -0,0 +1,161 @@
...
+            port: this.dbInstance.instanceEndpoint.port,
+            username: this.username,
+            passwordSecret: this.passwordSecretName,
```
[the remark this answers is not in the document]
t-richard: I think I misunderstood what you did here.

The idea was to fetch the secret from the application code, not via serverless framework integration with secret manager.

Wouldn't that need extra permissions?

This also comes with a performance penalty because it has to be fetched everytime.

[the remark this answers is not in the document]
mnapoli: Ah yes, yes and yes!

That's why there are 2 approaches:

1. deploy with the password in clear:
  - that's easier, simpler
  - but also less secure: the password is in clear in the Lambda console
  - and finally it prevents automatic rotation of DB passwords (because the Lambda function would need to be redeployed to get the new password)
  - free, with either environment variables, SSM parameter store, Serverless Dashboard secrets, etc.
2. deploy with the name of the secret: the application needs to retrieve the secret
  - more secure, as the password is decrypted at runtime
  - negative performance impact
  - benefits from password rotation
  - more expensive, as it requires Secrets Manager which has a fixed cost

That's why I think we need to provide support for both options.

src/constructs/aws/DatabaseSql.ts
mnapoli: Agreed, it wouldn't cost much I think.

mnapoli:
> I'll have to test again but it may be a problem if the user has no control over it.

Interesting!

Just to be clear:

- do you mean to allow users to override the version in the YAML config (easy)
- or do you mean that we must force users to always provide a precise version number in YAML? (more inconvenient)

piolet:
it could be interesting to have also the possibility to set parameters in the constructor, to create an `AWS::RDS::DBParameterGroup`. Something like :

```
constructs:
    database:
        type: database/mysql/dev
        password: qwertyqwerty
        parameters:
                - time_zone: Europe/Paris
                - max_connections: 600
```
instead of classic cloud formation :
```
RDSDBParameterGroup: 
  Type: 'AWS::RDS::DBParameterGroup' 
  Properties: 
    Description: CloudFormation Sample MySQL Parameter 
    Group Family: mysql8.0 
    Parameters: 
      time_zone: Europe/Paris 
      max_connections: 600 #'{DBInstanceClassMemory*3/4}'
```

What do you think? Maybe you have already thought about it?

Maybe could you expose the DbInstanceClassMemory for dynamic settings

metaskills:
Perhaps with the GA release of Aurora Serverless v2 this construct could use that instead?

t-richard:
Cloudformation support was not part of the GA release and so CDK support is not there yet.

We'll have to wait until support is available but it would be nice to have for sure.

There is also a different starting price so it might be interesting to have both standard RDS and Aurora serverless.

CSFlorin:
Cloudformation now supports Aurora Serverless v2 btw https://beabetterdev.com/2022/10/06/amazon-aurora-serverless-v2-cloudformation-example/ Hopefully the great work you've done so far can move forward now!

Issue #84, opened 2021-08-06
[Static Website] Integrate automatic domain DNS mapping within Route53


devsdevsdevs:
I am currently using the static website Serverless component found at https://github.com/serverless-components/website. It automatically sets up custom domain mappings with the Cloudfront CNAME if you are using Route53. Would it be possible to enable this feature here? I want to migrate to Lift as the Serverless component framework is restrictive in that I cannot use additional plugins etc.

mnapoli:
Continuing the discussion from #92.

One challenge is helping setting up the ACM certificate. Route53 can then be set up using CDK/CloudFormation AFAICT.

The certificate could be created via SDK calls. Here is an idea how, as a user, this could work:

```sh
$ npm i serverless-lift


# edit serverless.yml to add website with custom domain


$ serverless deploy
...
Error: The 'landing' website uses a custom domain (mywebsite.com), but no HTTPS certificate was found for that domain in ACM (in us-east-1).
Run 'serverless landing:certificate' to set up that certificate interactively.


$ serverless landing:certificate

# if no certificate exists
No HTTPS certificate matching 'mywebsite.com' was found in ACM in us-east-1.
To create a certificate manually in the AWS console, stop this command and follow that link: https://...
We can also create that certificate interactively:
How do you want to validate the domain?
- Email validation (...)
- DNS validation (...)
OK, creating the certificate.
The certificate has been created. You now need to validate it by email/DNS.
Once validated, you can run 'serverless deploy' and the certificate will automatically be used.
# exit

# if a certificate exists but isn't validated
A matching HTTPS certificate exists in ACM in the us-east-1 region: mywebsite.com / *.mywebsite.com
However, that certificate is not validated yet.
More details in the AWS Console: https://...
# exit

# if a certificate exists and is validated
A matching HTTPS certificate exists in ACM in the us-east-1 region: mywebsite.com / *.mywebsite.com
This certificate will automatically be used by Lift.
```

WDYT?

Is there any way to make it simpler with less steps?

Note: I'm afraid of adding interactive steps in `serverless deploy` directly, I don't really want to hijack the default behavior.

bobwallis:
For anyone coming across this... This is achievable using Lift plus only one extra plugin.

Certificate creation can be handled using https://www.serverless.com/plugins/serverless-certificate-creator

You can then access the certificate ARN in the Lift configuration like this:
```
constructs:
  website:
    type: server-side-website
    domain: ${env:DOMAIN}
[...]

PR #92, opened 2021-08-31
Add cname variable for static-website construct

fredericbarthelet:
When provisioning Route53 records using Cloudformation, there is no way to reference distribution CNAME generated by CloudFront (even while using custom domain names).

```yml
resources:
  Resources:
    Route53Record:
      Type: AWS::Route53::RecordSet
      Properties:
        HostedZoneId: XXXXX
        Name: app.mydomain.
        Type: A
        AliasTarget:
          HostedZoneId: XXXXX
          DNSName: !GetAtt ???.DomainName
```

This PR adds the following variable on the `static-website` construct :

- `cname`: the domain name of the resource, such as `d111111abcdef8.cloudfront.net`

### DIFF
--- a/docs/static-website.md
+++ b/docs/static-website.md
@@ -101,6 +101,35 @@ serverless landing:upload
 
 This command only takes seconds: it directly uploads files to S3 and clears the CloudFront cache.
 
+## Variables
+
+All static-website constructs expose the following variables:
+
+- `cname`: the domain name of the resource, such as `d111111abcdef8.cloudfront.net`
+
+This can be used to reference the bucket from Route53 configuration, for example:
+
+```yaml
+constructs:
+    landing:
+        type: static-website
+        path: public
+
+resources:
+  Resources:
+    Route53Record:
+      Type: AWS::Route53::RecordSet
+      Properties:
+        HostedZoneId: ZXXXXXXXXXXXXXXXXXXJ
+        Name: app.mydomain.
+        Type: A
+        AliasTarget:
+          HostedZoneId: ZXXXXXXXXXXXXXXXXXXJ
+          DNSName: ${construct:landing.cname}
+```
+
+_How it works: the `${construct:landing.cname}` variable will automatically be replaced with a CloudFormation reference to the CloudFront Distribution._
+
 ## Configuration reference
 
 ### Path
--- a/docs/static-website.md
+++ b/docs/static-website.md
@@ -120,11 +120,11 @@ resources:
     Route53Record:
       Type: AWS::Route53::RecordSet
       Properties:
-        HostedZoneId: ZXXXXXXXXXXXXXXXXXXJ
-        Name: app.mydomain.
+        HostedZoneId: ZXXXXXXXXXXXXXXXXXXJ # Your HostedZoneId
+        Name: app.mydomain
         Type: A
         AliasTarget:
-          HostedZoneId: ZXXXXXXXXXXXXXXXXXXJ
+          HostedZoneId: Z2FDTNDATAQYW2 # Cloudfront Route53 HostedZoneId. This does not change.
           DNSName: ${construct:landing.cname}
 ```
 
--- a/src/constructs/aws/StaticWebsite.ts
+++ b/src/constructs/aws/StaticWebsite.ts
@@ -71,6 +71,7 @@ export class StaticWebsite extends AwsConstruct {
         },
     };
 
+    private readonly distribution: Distribution;
     private readonly bucketNameOutput: CfnOutput;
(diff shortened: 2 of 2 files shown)


t-richard:
While having the cname as a variable makes sense, wouldn't it be better for this use case to have route 53 support inside the construct that would add a record to the hosted zone for each configured domain if a `hostedZone` parameter is provided?

docs/static-website.md
```diff
@@ -101,6 +101,35 @@ serverless landing:upload
...
+resources:
+  Resources:
+    Route53Record:
+      Type: AWS::Route53::RecordSet
+      Properties:
+        HostedZoneId: ZXXXXXXXXXXXXXXXXXXJ
+        Name: app.mydomain.
+        Type: A
+        AliasTarget:
+          HostedZoneId: ZXXXXXXXXXXXXXXXXXXJ
```
BaptistG, line 127: Maybe you could hardcode the HostedZoneId. The HostedZoneId here will always be the ID of the Route53 zone for Cloudfront (cloudfront.net) which never changes. The ID of this hosted zone is : Z2FDTNDATAQYW2

It will make it easier for someone who doesn't know Cloudfront very well to create the custom domain !

BaptistG:
@t-richard Indeed it would be even better if the plugin could handle creating the records when `hostedZone` is provided.

If `hostedZone` is provided the plugin could even handle certificate creation/validation  (would have to be created in us-east-1).

t-richard:
> the plugin could even handle certificate creation/validation (would have to be created in us-east-1).

The last part is the main problem. It would require creating and managing a whole new, separate stack just for the certificate and even then, manual action would be required to validate the certificate by DNS or email.

Whereas the route 53 part is quite easier if you have the hosted zone id.

fredericbarthelet:
> The last part is the main problem. It would require creating and managing a whole new, separate stack just for the certificate and even then, manual action would be required to validate the certificate by DNS or email.

Just checked with an AWS DA, you can actually bypass this problem using a custom resource from CloudFront, and provision the certificate using the SDK. It's not plain simple, but it does the job and is apparently widely used to solve this kind of issue. The [CDK offers the ability to easily implement such custom resource](https://docs.aws.amazon.com/cdk/api/latest/docs/custom-resources-readme.html), but it relies on having a bootstrap stack to provision corresponding lambda associated with the custom resource, which is not currently done in Lift.

As suggested @BaptistG merging as is (with the modification of the documentation) enables people to reference variable from the construct without impacting futur evolution of adding Route53 or not. I suggest we go ahead with this variable exposition and start discussion on bringing Routee53 in the construct in another PR.

docs/static-website.md
mnapoli: Sounds like a very good idea!

mnapoli (commented):
> As suggested @BaptistG, merging as is (with the modification of the documentation) enables people to reference variable from the construct without impacting futur evolution of adding Route53 or not. I suggest we go ahead with this variable exposition and start discussion on bringing Routee53 in the construct in another PR.

100% agree, it's a good small improvement, we can make it better in an another step.

The "HostedZoneId" comment is a good idea to change in the PR, and except that LGTM to merge and release!

docs/static-website.md
```diff
@@ -101,6 +101,35 @@ serverless landing:upload
...
+        type: static-website
+        path: public
+
+resources:
+  Resources:
+    Route53Record:
+      Type: AWS::Route53::RecordSet
+      Properties:
+        HostedZoneId: ZXXXXXXXXXXXXXXXXXXJ
+        Name: app.mydomain.
```
t-richard, line 124: I don't think this requires a trailing `.`
```suggestion
        Name: app.mydomain
```

t-richard (commented):
Expect this minor comment and the one from @BaptistG, it looks good to me !

t-richard:
I think for auto-setup on Route 53, we can wait for #44 to be merged and implement it for both at the same time.

> The CDK offers the ability to easily implement such custom resource, but it relies on having a bootstrap stack to provision corresponding lambda associated with the custom resource, which is not currently done in Lift.

CDK is filling the gap on that side (and other things like lookups, multi-stacks deploy, etc) but those require setting up an environment and sometimes this boostrap stack. This is clearly explained when using the CDK directly but might be less obvious when the CDK is shadowed like in Lift.

mnapoli:
Yeah, we can't really use CDK's custom resources because of the main flaw of Lift right now: we don't use the CDK's bootstrap stack. That means that we can't deploy any Lambda function with CDK.

Else is there any way we can do this more interactively? I've added a suggestion in #84 but any other idea is welcome! (and I think we should continue the discussion there)

Issue #94, opened 2021-09-07
server-side-website - missing default page for cdn


piolet:
In Server-side-website, we can configure the error page with :
```
constructs:
    website:
        # ...
        errorPage: error500.html
```
But we can't set the CDN's default page, like this :
```
constructs:
    website:
        # ...
        defaultPage: index.html
```

In my case, for example, index.html is added in my assets (it's a fully static page), but if i go to the page https://mydomain.com i redirect to the lambda mangement. If i go directly to https://mydomain.com/index.html, it's ok, but not really top.

What do you think ?

t-richard:
The server-side website is targeted at frameworks like Ruby on Rails, Symfony/Laravel, Django, Express, etc where the HTML rendering would be done server-side (with erb, twig/blade, DTL/Jinja, ejs/nunjucks, etc).

What you are willing to do looks like the [static website construct](https://github.com/getlift/lift/blob/master/docs/static-website.md), to run a static website generated by Jekyll or Hugo for example. Or Single page apps built with Vue.js or React.

Can you tell us a bit more about the use case here ?

piolet:
In my case, some page are static (full html) and some other are dynamic (php rendering).
All is ok, except for home page, my index.html. It's not a SPA.

t-richard:
This is somewhat out of scope of this construct (at least for now). 

**Maybe** you could trick the construct by doing something like

```yaml
constructs:
    website:
        type: server-side-website
        assets:
            '/css/*': public/css
            '/js/*': public/js
            '/': index.html
```

or

```yaml
constructs:
    website:
        type: server-side-website
        assets:
            '/css/*': public/css
            '/js/*': public/js
            '': index.html
```

This is from the top of my head, not sure it works but might be worth a try.

piolet:
with the first, I arrive on the list of the bucket :
```
<ListBucketResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
<Name>websiteassets2a73bb69-mhl8dw28rjm3</Name>
<Prefix/>
<Marker/>
<MaxKeys>1000</MaxKeys>
<IsTruncated>false</IsTruncated>
<Contents>
<Key>assets/css/styles.css</Key>
...
```

for the second, it's an error : 
`An error occurred: websiteCDN060D946D - Resource handler returned message: "Invalid request provided: The parameter globPattern is too small or empty. `

"flûte"

mnapoli:
[...]

PR #95, opened 2021-09-07
Add default root object parameter for cloud front

piolet:
to allow to manage a more accurate behavior of cloud front


piolet:
here is my proposal of evolution. This is my first fork on an open source project, you do what you want ;)

mnapoli:
Thanks!

TBH, I don't think it makes sense to add to the construct yet. As I mentioned in https://github.com/getlift/lift/issues/94#issuecomment-914316743 I'd rather wait to see if there's a significant demand for that, rather than take the risk to add complexity too early to the construct.

Let's reopen this pull request if we get more demand in #94!

piolet:
well, I hope to be able to use the custom constructs soon ;)

PR #97, opened 2021-09-12
feat: redirect to main domain in static and server-side websites

t-richard:
This PR implements redirection to a "main domain". This is especially useful for root domain to www redirections. 

The first domain will be used as the main domain where everything will redirect to if `redirectToMainDomain` is true. 

```yaml
constructs: 
    website: 
        type: "server-side-website"
        domain: 
            - www.example.com
            - example.com
        certificate: "arn:aws:acm:us-east-1:123456615250:certificate/0a28e63d-d3a9-4578-9f8b-14347bfe8123"
        redirectToMainDomain: true
```

This will redirect every request targeting `example.com` to `www.example.com`.

I've used Cloudfront functions to do this, it seems to be the best solution for this. There are other solutions with buckets doing redirects but this feels wrong to me now that we have Cloudfront Functions available.

### DIFF
--- a/docs/server-side-website.md
+++ b/docs/server-side-website.md
@@ -286,6 +286,22 @@ Usually, we can retrieve which domain a user is visiting via the `Host` HTTP hea
 
 The `server-side-website` construct offers a workaround: the `X-Forwarded-Host` header is automatically populated via CloudFront Functions. Code running on Lambda will be able to access the original `Host` header via this `X-Forwarded-Host` header.
 
+#### Redirect all domains to a single one
+
+It is sometimes necessary to redirect every request to a single domain. A common example is to redirect the root domain to the `www` version.
+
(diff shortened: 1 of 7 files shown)


mnapoli:
Awesome, that looks perfect to me! Thanks for that!

One last thing missing would be to add that to the documentation?

t-richard:
Yeah sure, I was just waiting for some feedback before comitting some time to write the docs :slightly_smiling_face:

t-richard:
I've added the docs

mnapoli:
Thank you!

mnapoli:
I just realized that the CloudFront function for "request" has been added for all websites, even when not useful.

This is a price and performance penalty (even though small) that we could skip, instead of deploying an empty function?

PR #100, opened 2021-09-20
React construct

mnapoli:
This PR adds a `react` website construct.

This construct is heavily based on the `static-website` construct.

Why a separate React construct? Yes, it is possible to deploy a React application with the existing [static website construct](https://github.com/getlift/lift/blob/master/docs/static-website.md). That new construct is preconfigured specifically for React. It is slightly more practical.

The difference will not especially be in technical terms. It is an attempt at being even closer to the real use cases of developers. In terms of adoption, we also want to see if we can get more attention from the React community by adapting serverless to _their_ world, instead of adapting React to the serverless world.

**In other words, starting from users and communities instead of starting from the technical solution.**

Example:

```yaml
service: my-app
provider:
    name: aws

constructs:
    website:
        type: react-website

plugins:
    - serverless-lift
```

On `serverless deploy`, the application will be built (via `npm run build`) and deployed as a public website.

[Full documentation](https://github.com/getlift/lift/blob/react/docs/react-website.md)

---

On a side note, this was an interesting first step into construct reuse (which could be big with #23).

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,9 +55,21 @@ The [AWS CDK](https://docs.aws.amazon.com/cdk/latest/guide/home.html) is extreme
 
 Lift provides components, aka "**constructs**", specifically selected for serverless applications. They are all built using the CDK and its best practices, with unique features to provide an awesome developer experience.
 
+### [React website](docs/react-website.md)
+
+Deploy React websites.
+
+```yaml
+constructs:
+    landing:
+        type: react-website
+```
+
+[Read more...](docs/react-website.md)
+
 ### [Static website](docs/static-website.md)
 
-Deploy static websites and single-page applications, for example React, VueJS or Angular apps.
+Deploy static websites and single-page applications, for example VueJS or Angular apps.
 
 ```yaml
 constructs:
--- a/docs/react-website.md
+++ b/docs/react-website.md
@@ -1,0 +1,126 @@
+# React website
+
+The `react-website` construct deploys React websites created with [Create React App](https://create-react-app.dev/).
+
+## Quick start
+
+```bash
+serverless plugin install -n serverless-lift
+```
+
+```yaml
+service: my-app
+provider:
+    name: aws
+
+constructs:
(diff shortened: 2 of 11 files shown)


piolet:
It also works for VueJS ou Angular ?
it is not specific to React a priori?
Maybe rename the builder to a more general purpose one?

mnapoli:
@piolet we already have `static-website` for that. It is specific to React.

piolet:
@mnapoli I don't understand your answer.
If there is already static-website for VueJS or Angular, why make a specific one for React ?

The difference is that the `npm run build` command is launched before the upload on S3 ? Well, this command also works for VueJS or Angular if I'm not wrong.

So why make a specific construct for React when it can also be applied to another framework?

mnapoli:
@piolet I answered that in the pull request description, is it still confusing to you?

piolet:
my remark is right: why name this construct "react-website" when without any other modification, it could also apply a VueJS or Angular application.

Or I didn't understand the description actually.

Tell me if I'm wrong, but the idea of this construct is to launch the build command, then get the dist directory to send it to S3, right?

mnapoli:
For VueJS the path of files to upload will be different. Also the dev commands might be different.

But also, from the description:

> The difference will not especially be in technical terms. It is an attempt at being even closer to the real use cases of developers. In terms of adoption, we also want to see if we can get more attention from the React community by adapting serverless to their world, instead of adapting React to the serverless world.
>
> In other words, starting from users and communities instead of starting from the technical solution.

mnapoli:
> I think the DX is nice but it kind of feels like trying to solve with code what may be solved with documentation.

That is a very very good point… I'll ponder on this. So far I haven't seen much enthusiasm about that approach, both internally and externally, so I'm pausing on this and thinking a bit more. Your idea may be a good approach 🤔

mnapoli:
Closing for now, I don't think this makes sense anymore.

PR #103, opened 2021-09-26
Add Upload construct

t-richard:
See the [docs](https://github.com/t-richard/lift/blob/upload-construct/docs/upload.md) for explanations.

- [ ] More tests
- [ ] Axios example ?
- [ ] in depth explanation of wha the function does ?
- [ ] example code for backend processing ?


src/constructs/aws/Upload.ts
```diff
@@ -0,0 +1,179 @@
...
+        };
+
+        this.bucket = new Bucket(this, "Bucket", {
+            encryption: encryptionOptions[resolvedConfiguration.encryption],
+            blockPublicAccess: BlockPublicAccess.BLOCK_ALL,
+            enforceSSL: true,
+            cors: [
+                {
+                    allowedMethods: [HttpMethods.PUT],
+                    allowedOrigins: ["*"],
```
t-richard, line 63: for this first version I chose to set wide CORS permissions. We may add an option to configure it later.

src/constructs/aws/Upload.ts
```diff
@@ -0,0 +1,179 @@
...
+            new PolicyStatement(
+                ["s3:PutObject", "s3:GetObject", "s3:DeleteObject", "s3:ListBucket"],
+                [this.bucket.bucketArn, Stack.of(this).resolve(Fn.join("/", [this.bucket.bucketArn, "*"]))]
+            ),
+        ];
+    }
+
+    outputs(): Record<string, () => Promise<string | undefined>> {
+        return {
+            bucketName: () => this.getBucketName(),
```
t-richard, line 141: it would be nice to output the full upload URL but I wasn't able to generate it properly for Rest APIs. Another problem was that people may have domains configured on their API Gateway or they may be using the server-side construct which woul make the outputed URL useless.

t-richard:
See some working examples here https://github.com/t-richard/lift-upload-example

Steps to deploy it are in the README

tschoffelen:
This is really cool!

Issue #111, opened 2021-10-05
Static website: Deployment IAM permissions


ondrejmirtes:
Hi,
I really appreciate how Lift helps me to remove some repeating work when deploying a static website, but one chore still remains - creating an AWS IAM user with the right permissions to deploy the site from my CI/CD pipeline.

What I usually do is a trial-and-error procedure of creating a user and adding more and more permissions until my deployment is successful. Is there an easier way? Is it documented?

Thank you!

t-richard:
I've went through the same process and can't agree more. 

IAM is complex and IMO Serverless Framework does a poor job at pointing you in the right direction (or I'm not aware of it). They even recommend creating a user with `Administrator Access` in there [docs](https://www.serverless.com/framework/docs/providers/aws/guide/credentials/#creating-aws-access-keys).

I'm not sure what's the best practise in this field right now but what I've done so far is generating a user with the least permissions I can come up with. And yes this is trial & error each time. 

I also know you can make Cloudformation assume a role when deploying but that wouldn't help with your static files upload problem here.

I think the required permissions to use and deploy a construct could be documented but there would still be a missing piece in Serverless Framework.

It would be nice for serverless framework to have a command that looks at what would be deployed and generates the proper IAM permission document for deployment (and them sending an event so that Lift can hook into the process and add its own to the mix).

Hope that makes sense. Would love to hear thoughts on this.

mnapoli:
Yes, yes and yes.

We've discussed it several times internally at Serverless, and it's a tough problem to solve. No tools, as far as we know, has a magic solution for this.

One idea we may want to explore is:

- let each construct expose the permissions it requires
- then there would be a command that lets user export those permissions (useful for reviewing as well), even before deploying
- on deployment, it could even be great if Lift/Serverless Framework would check the permissions before deploying to anticipate any failure

But, as you can guess, this is a lot of work. I'm curious what you think about that approach, or if you have anything else in mind.

For those stumbling on this discussion feel free to add a 👍  to the issue: that helps us gauge interest and prioritize the roadmap.

PR #115, opened 2021-10-08
[static-website] fix: stop silent fail on s3 deleteObjects

fargito:
Hello and thanks a lot for this awesome project!

When using the `static-website` construct on my project, I realized that I had forgotten the `s3:deleteObject` IAM permission of my deployment policy. Therefore, the s3 files synced with lift were never deleted. However, no error was thrown and the deployement failed silently.

This silent fail is due to a particularity in the S3 `deleteObjects` [SDK](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#deleteObjects-property) and the [corresponding HTTP API](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html). Since this API can encounter partial failures, the call never fails but returns the errors in a `Errors` key.

I think that Lift should explicitely fail when the `Errors` key is defined and the list of errors is not empty. In this case, the proposed error is the following:

```
Deleting index.html
{
  Key: 'index.html',
  Code: 'AccessDenied',
  Message: 'Access Denied'
}
 
 Serverless Error ----------------------------------------
 
  Unable to delete files
 
  Get Support --------------------------------------------
     Docs:          docs.serverless.com
     Bugs:          github.com/serverless/serverless/issues
     Issues:        forum.serverless.com
 
  Your Environment Information ---------------------------
     Operating System:          linux
     Node Version:              14.16.1
     Framework Version:         2.61.0 (local)
     Plugin Version:            5.4.5
     SDK Version:               4.3.0
     Components Version:        3.17.1
 
```

### DIFF
--- a/src/utils/s3-sync.ts
+++ b/src/utils/s3-sync.ts
@@ -15,6 +15,7 @@ import { lookup } from "mime-types";
 import { chunk, flatten } from "lodash";
 import chalk from "chalk";
 import type { AwsProvider } from "@lift/providers";
+import ServerlessError from "./error";
 
 const readdir = util.promisify(fs.readdir);
 const stat = util.promisify(fs.stat);
@@ -147,7 +148,7 @@ export async function s3Put(aws: AwsProvider, bucket: string, key: string, fileC
 }
 
 async function s3Delete(aws: AwsProvider, bucket: string, keys: string[]): Promise<void> {
-    await aws.request<DeleteObjectsRequest, DeleteObjectsOutput>("S3", "deleteObjects", {
+    const response = await aws.request<DeleteObjectsRequest, DeleteObjectsOutput>("S3", "deleteObjects", {
         Bucket: bucket,
         Delete: {
             Objects: keys.map((key) => {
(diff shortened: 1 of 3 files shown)


t-richard:
Also kinda linked to #111 because this originated from misconfigured deploy credentials

t-richard:
I also think failing is somewhat better than a warning because it could lead to unnecessary objects being stored and billed which could become a significant cost over time.

> Concerning the explicit error, I logged the error returned by the S3 api (as in the description of the PR).

I know you logged the errors but what I was suggesting is to add instructions on **how to solve the issue**

Trying to give users some guidance on what may be wrong will help to get a good developer experience and reduce frustration (and IAM is a very good frustration generator :sweat_smile:).

See some examples:
* https://github.com/getlift/lift/blob/master/src/constructs/aws/StaticWebsite.ts#L187-L190
* https://github.com/getlift/lift/blob/master/src/constructs/aws/StaticWebsite.ts#L86-L90

fargito:
@t-richard done :)

fargito:
@mnapoli actually it is also used by the "server-side-website" contruct so I updated the tests and added it to the error message

PR #118, opened 2021-10-11
Drop NodeJS 10 support

t-richard:


t-richard:
Had to start at node 12.13 because `constructs` requires at least node 12.7 and `jest` requires at least node 12.13

But pest is a dev dependency so not sure that's the way to do it.

I wanted to add a check in CI for the lowest node version supported by I guess it's not the way to do it. Any suggestion appreciated.

PR #120, opened 2021-10-12
Support delay on queues

### LINKED ISSUE #119: [Queue] Allow FIFO queue construct to declare delivery delay
Hi, (please do point me in the right direction if it's already possible)

is there a possibility of programatically setting the [delivery delay](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-delay-queues.html) in the serverless.yml declaration?

t-richard:
Would fix #119 

I've worked on this because the change is small enough. 

If you're ok to merge it, I'll write the docs.

### DIFF
--- a/docs/queue.md
+++ b/docs/queue.md
@@ -299,6 +299,24 @@ constructs:
             # The retry delay on the queue will be 10*6 => 60 seconds
 ```
 
+### Delivery delay
+
+When a message is sent to the queue, it will be available immediately to the worker.
+
+You can postpone the delivery of messages by a given amount of seconds using the `delay` option.
+
+The maximum value is 900 seconds (15 minutes).
+
+```yaml
+constructs:
+    my-queue:
+        # ...
+        worker:
+            handler: src/worker.handler
+            # Messages delivery will be delayed by 1 minute
+            delay: 60
+```
+
 ### Batch size
 
 ```yaml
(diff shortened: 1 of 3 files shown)


mnapoli:
Thank you!

Senseye:
I tried to add `delay: 60` to my queue config like this:

```
constructs:
  aswesome-sqs-queue:
    type: queue
    worker:
      handler: src/handlers/awesome.handler
      delay: 60
```

Now I expect `Delivery delay` to be set in Amazon SQS; but unfortunately it remains `0 Seconds`:
image

Am I doing something wrong?

t-richard:
@Senseye it should probably be

```
constructs:
  aswesome-sqs-queue:
    type: queue
    delay: 60
    worker:
      handler: src/handlers/awesome.handler
```

Let me know if that solves the issue.

Senseye:
> @Senseye it should probably be
> 
> ```
> constructs:
>   aswesome-sqs-queue:
>     type: queue
>     delay: 60
>     worker:
>       handler: src/handlers/awesome.handler
> ```
> 
> Let me know if that solves the issue.

@t-richard thought so! 

Then this readme needs to be updated:
https://github.com/getlift/lift/blob/master/docs/queue.md#delivery-delay
The example there is incorrect.

Thanks Richard!

mnapoli:
Thanks! That's fixed in e3f8f1cbb15480803bb72a48f6b3c2a8a37b13a7

PR #131, opened 2021-11-05
Support Serverless Framework v3 & integrate with the new design

mnapoli:
Serverless Framework v3 will come with a new CLI design, and new plugin APIs to integrate with the new design.

This pull request adds support for the v3 beta (announced soon), but keeps compatibility with v2. The reason is to avoid having to publish multiple releases to support v2 and v3 (and the nightmare for users).

That leads to sometimes verbose code :/ Not ideal, but at least the experience for users should be great.

Below are some screenshots ✨ 
Capture d’écran 2021-11-05 à 17 09 12

https://user-images.githubusercontent.com/720328/140542812-144a538a-8c90-405f-81b6-4936a4e77323.mp4

Screen 2021-11-05 17-35

https://user-images.githubusercontent.com/720328/140542827-1d7ed102-5808-4898-8962-69c650e44cce.mp4

https://user-images.githubusercontent.com/720328/140542832-c91111f1-5705-47a9-b076-db8557cf0dbf.mp4

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,6 +92,6 @@
   },
   "types": "dist/src/plugin.d.ts",
   "peerDependencies": {
-    "serverless": "^2.36.0 || pre-3"
+    "serverless": "^2.36.0 || ^3"
   }
 }
--- a/package.json
+++ b/package.json
@@ -92,6 +92,6 @@
   },
   "types": "dist/src/plugin.d.ts",
   "peerDependencies": {
-    "serverless": "^2.36.0"
+    "serverless": "^2.36.0 || pre-3"
   }
 }
--- a/src/constructs/aws/Queue.ts
+++ b/src/constructs/aws/Queue.ts
@@ -7,6 +7,7 @@ import type { Construct as CdkConstruct } from "@aws-cdk/core";
 import { CfnOutput, Duration } from "@aws-cdk/core";
 import chalk from "chalk";
 import type { PurgeQueueRequest, SendMessageRequest } from "aws-sdk/clients/sqs";
+import type { Ora } from "ora";
 import ora from "ora";
 import { spawnSync } from "child_process";
 import * as inquirer from "inquirer";
@@ -18,6 +19,8 @@ import { sleep } from "../../utils/sleep";
 import { PolicyStatement } from "../../CloudFormation";
 import type { CliOptions } from "../../types/serverless";
 import ServerlessError from "../../utils/error";
+import type { Progress } from "../../utils/logger";
+import { getUtils } from "../../utils/logger";
 
 const QUEUE_DEFINITION = {
     type: "object",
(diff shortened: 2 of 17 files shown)


package.json
```diff
@@ -20,6 +20,7 @@
     "@aws-cdk/aws-sns": "^1.110.1",
     "@aws-cdk/aws-sqs": "^1.110.1",
     "@aws-cdk/core": "^1.110.1",
+    "@serverless/utils": "^5.20.0",
```
mnapoli, line 23: TODO update to `pre-6`?

package.json
```diff
@@ -92,6 +93,6 @@
   },
   "types": "dist/src/plugin.d.ts",
   "peerDependencies": {
-    "serverless": "^2.36.0"
+    "serverless": "^2.36.0 || pre-3"
```
mnapoli, line 96: We may be able to remove that: to test

src/constructs/aws/Queue.ts
```diff
@@ -7,17 +7,21 @@ import type { Construct as CdkConstruct } from "@aws-cdk/core";
 import { CfnOutput, Duration } from "@aws-cdk/core";
 import chalk from "chalk";
 import type { PurgeQueueRequest, SendMessageRequest } from "aws-sdk/clients/sqs";
+import type { Ora } from "ora";
```
fredericbarthelet, line 10: Can you add dev dependency `ora` as well in package ?

src/constructs/aws/Queue.ts
```diff
@@ -250,53 +254,93 @@ export class Queue extends AwsConstruct {
...
+        let progressV2: Ora | undefined;
+        let progressV3: Progress | undefined;
+        if (isV3) {
+            progressV3 = progress.create({
+                message: "Polling failed messages from the dead letter queue",
+            });
+        } else {
+            progressV2 = ora("Polling failed messages from the dead letter queue").start();
+        }
         const messages = await pollMessages({
```
fredericbarthelet, line 271: maybe for readability, it would be easier to have 2 callbacks. One for v2 and one for v3

src/plugin.ts
```diff
@@ -440,16 +462,27 @@ class LiftPlugin {
...
 
     private async eject() {
-        log("Ejecting from Lift to CloudFormation");
+        if (isV3) {
+            log("Ejecting from Lift to CloudFormation");
+            log();
+        } else {
+            legacyLog("Ejecting from Lift to CloudFormation");
+            legacyLog("");
+        }
```
fredericbarthelet, line 471:
```suggestion
        const logger = isV3 ? log : legacyLog;
        logger("Ejecting from Lift to CloudFormation");
```

src/utils/logger.ts
```diff
@@ -1,11 +1,22 @@
 import chalk from "chalk";
+import { log } from "@serverless/utils/log";
 
-export function log(message: string): void {
+export let isV3 = false;
```
fredericbarthelet, line 4: We could contain switch between v2 and v3 in terms of logging to this file in order not to "pollute" all constructs with the implementation variations between the versions of serverless.

We could have in this logger file an API similar to v3 (with all log level being different method) and call v3 API or v2 unique way of logging (hereby `console.log`) depending on framework version. We would have to replace `log` calls to this file in code base with corresponding required log level.

WDYT ?

mnapoli: 👍  done that. It was easily possible except for progresses (where the API difference between Ora and the serverless progress is very different)

src/constructs/aws/Queue.ts
mnapoli: I've gone over the whole PR to try and minimize the differences, but in some cases honestly it wasn't easy to do much better.

PR #135, opened 2021-11-18
use path.posix.join() to avoid issues with OS path separators that co…

BenMenking:
On the Windows platform using Cygwin64, path.join() appears to use the Windows path separator ('\') instead of the POSIX path separator which causes files in directories uploaded to AWS S3 to contain a backslash and become inaccessible.  path.posix.join() appears to use the correct path separator for AWS S3.

Example of issue 
---
serverless.yml config:
```
    assets:
      '/js/*': assets/js
```
`aws s3 ls <bucket>` listing:
```
2021-11-18 15:29:18      50205 js\analytics.js
2021-11-18 15:29:18       8233 js\bootstrap-datepicker.min.js
2021-11-18 15:29:19      36399 js\bootstrap.min.js
```
After code changes, `aws s3 ls <bucket>`:
```
                           PRE js/
```
And `aws s3 ls <bucket>/js/`:
```
2021-11-18 20:57:23      50205 analytics.js
2021-11-18 20:57:23       8233 bootstrap-datepicker.min.js
2021-11-18 20:57:23      36399 bootstrap.min.js
```

hth, use as needed

### DIFF
--- a/src/constructs/aws/ServerSideWebsite.ts
+++ b/src/constructs/aws/ServerSideWebsite.ts
@@ -256,7 +256,7 @@ export class ServerSideWebsite extends AwsConstruct {
                 invalidate = invalidate || hasChanges;
             } else {
                 // File
-                const targetKey = path.join(s3PathPrefix, path.basename(filePath));
+                const targetKey = path.posix.join(s3PathPrefix, path.basename(filePath));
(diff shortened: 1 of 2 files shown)


mnapoli:
Thank you!

PR #141, opened 2021-11-27
Issue templates forms

t-richard:
I noticed we tend to regularly ask people the same information when they create an issue, especially when requesting a new feature (what is the use-case, what config could look like, etc).

This PR introduces issue templates **forms**

This is a recent Github feature that creates an actual form which users have to fill. 

I tried to make it simple enough so that is doesn't discourages people from submitting but still give a bit more structure.

You can see a preview on my fork, feel free to create an issue there to see how it looks https://github.com/t-richard/lift/issues/new/choose

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -3,11 +3,6 @@ description: Report a bug in Lift
 labels: bug
 
 body:
-    - type: input
-      id: construct
-      attributes:
-          label: Affected Construct
-          placeholder: If applicable (eg. queue, storage)
     - type: textarea
       id: description
       attributes:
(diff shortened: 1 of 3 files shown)


mnapoli:
Looks really good to me, thanks for doing that!

The only thing I would improve is try to fit the "bug report" form with 3 textarea (instead of 4). That's an arbitrary number, but currently the form feels a bit too much given we have a low volume of issues, sometimes bug reports are really obvious they don't need so many fields (even optional).

t-richard:
You're right :+1: 

I've done the following
* remove the affected construct, this should be obvious in most cases
* remove the "possible solution" block and suggest to use the "additional information" one instead

Let me know if that sounds better!

mnapoli:
LGTM, thanks for the update!

Let's give this a try and adjust if needed.

PR #145, opened 2021-12-03
[QUEUE] Support partial batch failure

t-richard:
Support for `functionResponseTypes` in Serverless framework was added in https://github.com/serverless/serverless/pull/10297

Because Lift registers the SQS event mapping automatically, we need to expose this option too.

As we try to be focused on use-case instead of matching the AWS naming, I went with a boolean `partialBatchFailure` config option (see the docs for usage). No hard opinion on this name, feel free to suggest something else.

- [x] Tests
- [x] Documentation

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
                     - 14
                     - 16
                 serverless:
-                    - "2.67.0"
+                    - "2.36.0"
                     - latest
         steps:
             - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
                     - 14
                     - 16
                 serverless:
-                    - "2.36.0"
+                    - "2.67.0"
                     - latest
         steps:
             - uses: actions/checkout@v2
--- a/docs/queue.md
+++ b/docs/queue.md
@@ -45,6 +45,7 @@ Lift constructs are production-ready:
 - Failed messages in the dead letter queue are stored for 14 days (the maximum) to give developers time to deal with them
 - The SQS "Visibility Timeout" setting is configured per AWS recommendations ([more details](#retry-delay))
 - Batch processing is disabled by default ([configurable](#batch-size)): errors need to be handled properly using [partial batch failures](#partial-batch-failures)
+- The event mapping is configured with `ReportBatchItemFailures` enabled by default for [partial batch failures](#partial-batch-failures) to work out of the box
 
 ## Example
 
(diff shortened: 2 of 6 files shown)


fredericbarthelet:
Hey @t-richard, thanks for your contribution !
After discussing it with @mnapoli, we'd like to suggest the following :

- There currently is a significant amount of developers using Lift with serverless framework version lower than 2.67. Introducing such requirement to migrate up to v2.67 would require a lot of people to manually upgrade framework version (the framework is listed as peer dependency, the behavior of updating depends on npm version).
- There is no downside in using `ReportBatchItemFailures` for the event source mapping by default. People not using this feature can continue to use the same code base in their lambda to process items before the feature was implemented

Considering the above, would you be open to update your PR with the following recommendations:

- Keep minimum serverless framework version at 2.36.0
- Remove the `partialBatchFailure` option of the Queue construct
- Always set `responseType` to `ReportBatchItemFailures` in the construct for the lambda configuration input
- Update Queue documentation to state that event source mapping will always declare `ReportBatchItemFailures` configuration
- Add a warning in the Queue documentation to state that partial batch failure only works if serverless framework version is higher than 2.67

WDYT ?

t-richard:
@fredericbarthelet thanks for the feedback.

I agree that would be ideal that's what I first thought about.

> There is no downside in using ReportBatchItemFailures for the event source mapping by default. People not using this feature can continue to use the same code base in their lambda to process items before the feature was implemented

What made me doubt about this is [that part of the AWS documentation](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#services-sqs-batchfailurereporting), especially the first bullet point. 

> Lambda treats a batch as a complete failure if your function returns any of the following:
> 
> * An invalid JSON response
> * An empty string itemIdentifier
> * A null itemIdentifier
> * An itemIdentifier with a bad key name
> * An itemIdentifier value with a message ID that doesn't exist

We don't have control over what is done in the handler in userland so if they are returning something in their handler, it might fail messages for unexpected reasons.

We are probably talking about edge-cases here but I think it's worth dicussing. 

If you're OK with this, I'll happily update the PR :slightly_smiling_face:

mnapoli:
> We don't have control over what is done in the handler in userland so if they are returning something in their handler, it might fail messages for unexpected reasons.

That's a good point. But personally I think this is an acceptable compromise: before the new feature, returning anything from SQS handlers would make no sense. So 99% of the cases would work correctly:

image

t-richard:
Updated following the discussion above :slightly_smiling_face:

t-richard:
Fixed the tests. 

I'm no TypeScript pro so not sure if my solution is the correct one. I couldn't make it work to properly type `version` in the type definition.

mnapoli:
Thanks!

PR #152, opened 2022-01-06
Mark plugin as compatible with v3 of `serverless`

medikoo:
It's needed so users can upgrade to v3 of the Framework without issues

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,6 +92,6 @@
   },
   "types": "dist/src/plugin.d.ts",
   "peerDependencies": {
-    "serverless": "^2.36.0"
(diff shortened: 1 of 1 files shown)


t-richard:
I think it would be better to do this in https://github.com/getlift/lift/pull/131 and merge it.

medikoo:
@t-richard good point. If #131 will be merged on time, then this can safely be closed. 

Still, otherwise, this plugin will work without issues in v3m as is in master, so just taking this in and publishing it before v3 is released, will ensure that users have no issues in upgrading when v3 is released

mnapoli:
Thanks! Finally merged #131

Issue #160, opened 2022-02-01
Queue: Allow setting the maximumBatchingWindow


jakejscott:
### Start from the Use-case

Currently the `maximumBatchingWindow` is hardcoded to 60 seconds (when using standard queue). We would like to be able to configure this. There's even a TODO in the code for adding this as a setting :)

```
this.configuration.worker.events = [
            // Subscribe the worker to the SQS queue
            {
                sqs: {
                    arn: this.queue.queueArn,
                    batchSize: batchSize,
                    // TODO add setting
                    maximumBatchingWindow: this.queue.fifo ? undefined : 60,
                    functionResponseType: "ReportBatchItemFailures",
                },
            },
        ];
```

### Example Config

```yaml
constructs:
    my-queue:
        type: queue
        batchSize: 10
        maxBatchingWindow: 5 # <-- add this
        worker:
            handler: src/report-generator.handler
```

### Implementation Idea

NOTE: The `maxBatchingWindow` setting is in seconds

jakejscott:
Also in the docs it says:

> We recommend setting your queue visibility timeout to six times your function timeout, plus the value of MaximumBatchingWindowInSeconds

Currently I think the `visibilityTimeout` calculation is not taking into account the 60s default for maxBatchWindow.

jakejscott:
Also I'm wondering if the default for maxBatchWindow should be 0 rather than 60 as it might catch people out?

mnapoli:
Closed by #161

PR #161, opened 2022-02-01
Queue: Add setting for maximumBatchingWindow

jakejscott:
I added an issue for this here https://github.com/getlift/lift/issues/160

### DIFF
--- a/docs/queue.md
+++ b/docs/queue.md
@@ -335,6 +335,22 @@ Note you can use [partial batch failures](#partial-batch-failures) to avoid fail
 
 It is possible to set the batch size between 1 and 10.
 
+### Maximum Batching Window
+
+```yaml
+constructs:
+    my-queue:
+        # ...
+        maxBatchingWindow: 5 # SQS will wait 5 seconds (so that it can batch any requests together) before delivering to lambda
+```
+
+*Default: 60 seconds*
+
+The maximum amount of time to gather records before invoking the lambda. This increases the likelihood of a full batch at the cost of delayed processing.
+
+It is possible to set the `maxBatchingWindow` between 0 and 300. 
+
+
 ### Partial batch failures
 
 When using message batches, an error thrown in your worker function would consider the whole batch as failed.
@@ -356,6 +372,7 @@ It contains the identifier of the messages you consider as failed in the `itemId
 ```
 
 You can learn more in the [official AWS documentation](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#services-sqs-batchfailurereporting).
+
 
 ### More options
 
--- a/docs/queue.md
+++ b/docs/queue.md
@@ -341,7 +341,7 @@ It is possible to set the batch size between 1 and 10.
 constructs:
     my-queue:
         # ...
-        maxBatchingWindow: 5 # SQS will wait 5 seconds (so that it can batch any requests together) before delivering to lambda
+        maxBatchingWindow: 5 # SQS will wait 5 seconds (so that it can batch any messages together) before delivering to lambda
 ```
 
 *Default: 60 seconds*
--- a/docs/queue.md
+++ b/docs/queue.md
@@ -344,7 +344,7 @@ constructs:
         maxBatchingWindow: 5 # SQS will wait 5 seconds (so that it can batch any messages together) before delivering to lambda
 ```
 
-*Default: 60 seconds*
+*Default: 0 seconds*
 
 The maximum amount of time to gather records before invoking the lambda. This increases the likelihood of a full batch at the cost of delayed processing.
 
--- a/src/constructs/aws/Queue.ts
+++ b/src/constructs/aws/Queue.ts
@@ -53,8 +53,6 @@ const QUEUE_DEFINITION = {
 } as const;
 type Configuration = FromSchema<typeof QUEUE_DEFINITION>;
 
-const MAX_BATCHING_WINDOW = 60;
-
 export class Queue extends AwsConstruct {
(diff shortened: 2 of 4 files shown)


src/constructs/aws/Queue.ts
```diff
@@ -40,6 +40,11 @@ const QUEUE_DEFINITION = {
             minimum: 1,
             maximum: 10,
         },
+        maxBatchingWindow: {
```
t-richard, line 43: I'm not sure of the name, each time I have to search for it because it's not clear what it does just by reading the name.

We often try to think outside of the cloudformation/cdk names to find something meaningful if possible.

I unfortunately have no idea on a better name in this case 😅

t-richard (commented):
Thanks for submitting this 🙏

Just some minor comments or questions.

Documentation would also be appreciated 😁

src/constructs/aws/Queue.ts
```diff
@@ -240,6 +252,7 @@ export class Queue extends AwsConstruct {
     private appendFunctions(): void {
         // The default batch size is 1
         const batchSize = this.configuration.batchSize ?? 1;
+        const maximumBatchingWindow = this.queue.fifo ? 0 : this.configuration.maxBatchingWindow ?? MAX_BATCHING_WINDOW;
```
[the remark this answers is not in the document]
jakejscott: I can set it back to undefined no problem. 

The reason why I set it to 0 was from the docs it says that 0 is the default: https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#services-sqs-params

And TypeScript was complaining in that unary expression :)

src/constructs/aws/Queue.ts
jakejscott: maxBatchingWindowInSeconds?
batchDelay?

jakejscott: @t-richard I'll update the docs -- Just wasn't sure if you would be ok with the direction in terms of changing how the visibility timeout is calculated :)

docs/queue.md
```diff
@@ -335,6 +335,22 @@ Note you can use [partial batch failures](#partial-batch-failures) to avoid fail
...
+### Maximum Batching Window
+
+```yaml
+constructs:
+    my-queue:
+        # ...
+        maxBatchingWindow: 5 # SQS will wait 5 seconds (so that it can batch any messages together) before delivering to lambda
+```
+
+*Default: 60 seconds*
```
mnapoli, line 347: https://github.com/getlift/lift/issues/160#issuecomment-1027352596

> Also I'm wondering if the default for maxBatchWindow should be 0 rather than 60 as it might catch people out?

Actually I agree with you… CloudFormation says the default is `0` for SQS:

image

I would say we should switch to 0, WDYT @t-richard?

src/constructs/aws/Queue.ts
mnapoli: > I unfortunately have no idea on a better name in this case 😅

Same, I don't see a better alternative. `maxBatchingWindowInSeconds` is good but very long 🤷

docs/queue.md
t-richard: AFAIK the batching window is mostly for cost/usage optimization. It will wait longer so it can eventually gather more messages for a batch which will avoid too much Lambda invocations.

Also, setting this to something else than `0` allows for batches of more than 10 items which we don't currently allow (but maybe we should :slightly_smiling_face:).

There's no clear answer here, it depends on the use case people make from the queue. Sometimes performance will be prefered, sometimes usage optimization. 

If I had to make a choice, I'd keep it as is. 60 seconds seems like a good default to me because it allows for better optimization without impacting performance too much. It will have no impact for batchSize of 1 and most of the time you are seeking optimization when using batches.

src/constructs/aws/Queue.ts
t-richard: `batchWaitPeriod` ? Not sure about this one too but it keeps the discussion going :smile:

jakejscott: batchBuffer

docs/queue.md
mnapoli: Thanks, I wasn't sure if there were scenarios I was missing.

Let's use `0` as default then, because:

- it's consistent with the actual default from AWS
- it's not surprising (anything else is surprising because it makes the queue non-real-time, which isn't stated upfront in this construct)

As you mentioned, setting any other value is an optimization that users can make based on the specific usage for their queue, so let's keep it real-time by default.

src/constructs/aws/Queue.ts
mnapoli: Since we don't see an obvious alternative name, let's stick with `maxBatchingWindow` then.

Thanks for bringing this up though, this is an interesting exercise. It just doesn't always pay off  😅

docs/queue.md
jakejscott: I'll update the PR with `0` as the default.

luigidragone96:
Hello, is there any news regarding the release of this modification?
It would be very useful to be able to change the hard-coded value of 60 seconds.

predam:
Hello guys! I really need to change the value of maximumBatchingWindow , do you need any help in order to complete this merge?

src/constructs/aws/Queue.ts
predam: in aws docs the name for this prop is 'MaximumBatchingWindowInSeconds'

mnapoli (approved):

mnapoli:
Looking great, thank you!

PR #162, opened 2022-02-02
fix(queue): allow retry more than 10 jobs

iam4x:
## Issue

> Maximum number of entries per request are 10. You have sent 30.

image

## Fix

* AWS SQS allow batches of maximum 10 entries / messages
* When retrying messages in dlq we should chunk messages in batch of 10

## Output

image


mnapoli (approved):
Thank you that looks great!

Before I merge can you just confirm that this PR solved the bug for you?

iam4x:
Yes ser, if you have more than 10 failed jobs in the dlq the actual retry command won’t work.

mnapoli:
Thank you!

PR #165, opened 2022-02-06
Fix SPA error page behavior

### LINKED ISSUE #163: Static website deploy does not work for SPA projects anymore
### Description

I've been using lift for the past 6 months and it's a great plugin (thanks to all the maintainers).
After updating to lift 1.12.0 and serverless 3.1.0 I noticed a strange behaviour which appears to be related to one of aws-cdk subpackages (I can not pinpoint which one).
For some reason the policy of the bucket is not being set correctly anymore which causes cloudfront to return a 403 instead of rendering index.html when accessing any route (which is needed for SPA projects).

### How to Reproduce

Create a simple project with serverless
```shell
yarn init
yarn add -D serverless serverless-lift
mkdir dist
echo hello > dist/index.html
```

Put the following in serverless.yml
```yml
service: lift-test
frameworkVersion: '3'
useDotenv: true
variablesResolutionMode: 20210326

provider:
  name: aws
  region: eu-west-1
  stage: ${opt:stage}
  stackTags:
    STACK: ${self:service}
    REGION: ${self:provider.region}

constructs:
  vue:
    type: static-website
    path: dist

plugins:
  - serverless-lift
```
And run `yarn sls deploy --stage dev`
Once deploy is done you can see access the cloudfront domain (xyz.cloudfront.net/whatever) and you get 403 instead of hello

### Additional Information

From what I understand the issue is coming from incorrect permissions set on the bucket.

For the latest version which SPA does not work anymore
```yml
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::cloudfront:user/CloudFront Origin Access Identity XYZ"
            },
            "Action": "s3:GetObject",
            "Resource": "arn:aws:s3:::test-dev-vuebucketXYZ/*"
        }
    ]
}
```

For the old one that works
```yml
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::cloudfront:user/CloudFront Origin Access Identity XYZ"
            },
            "Action": [
                "s3:GetObject*",
                "s3:GetBucket*",
                "s3:List*"
            ],
            "Resource": [
                "arn:aws:s3:::trope3-webapp-dev-vuebucketXYZ",
                "arn:aws:s3:::trope3-webapp-dev-vuebucketXYZ/*"
            ]
        },
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::cloudfront:user/CloudFront Origin Access Identity XYZ"
            },
            "Action": "s3:GetObject",
            "Resource": "arn:aws:s3:::trope3-webapp-dev-vuebucketXYZ/*"
        }
    ]
}

### PR DESCRIPTION
**Pull request by adriencaccia**
Switch the origin bucket configuration to static website hosting,
leaving the configuration on the bucket side, without having to rely
on a custom origin access identity.

Fix #163, #164

An other possible solution would be to simply re-attach a custom origin access policy with the  `s3::ListBucket` policy, as explained in [my comment](https://github.com/getlift/lift/issues/163#issuecomment-1030204356) in #163. The reason I did not implement this solution is to match lift goal on leveraging as much CDK default behaviors as possible.
If this solution is preferred, I would be happy to update the PR :slightly_smiling_face:

### DIFF
--- a/src/constructs/aws/StaticWebsite.ts
+++ b/src/constructs/aws/StaticWebsite.ts
@@ -9,13 +9,12 @@ import {
 } from "@aws-cdk/aws-cloudfront";
 import * as cloudfront from "@aws-cdk/aws-cloudfront";
 import type { Construct as CdkConstruct } from "@aws-cdk/core";
-import { CfnOutput, Duration, RemovalPolicy } from "@aws-cdk/core";
+import { CfnOutput, RemovalPolicy } from "@aws-cdk/core";
 import type { FromSchema } from "json-schema-to-ts";
 import chalk from "chalk";
 import { S3Origin } from "@aws-cdk/aws-cloudfront-origins";
 import * as acm from "@aws-cdk/aws-certificatemanager";
 import { flatten } from "lodash";
-import type { ErrorResponse } from "@aws-cdk/aws-cloudfront/lib/distribution";
 import type { AwsProvider } from "@lift/providers";
 import { AwsConstruct } from "@lift/constructs/abstracts";
 import type { ConstructCommands } from "@lift/constructs";
@@ -91,6 +90,11 @@ export class StaticWebsite extends AwsConstruct {
         }
 
         const bucket = new Bucket(this, "Bucket", {
+            // Enable static website hosting
+            websiteIndexDocument: "index.html",
+            websiteErrorDocument: this.errorResponseDocument(),
+            // Required when static website hosting is enabled
+            publicReadAccess: true,
             // For a static website, the content is code that should be versioned elsewhere
             removalPolicy: RemovalPolicy.DESTROY,
         });
@@ -120,8 +124,6 @@ export class StaticWebsite extends AwsConstruct {
 
         this.distribution = new Distribution(this, "CDN", {
             comment: `${provider.stackName} ${id} website CDN`,
-            // Send all page requests to index.html
-            defaultRootObject: "index.html",
             defaultBehavior: {
                 // Origins are where CloudFront fetches content
                 origin: new S3Origin(bucket),
@@ -132,7 +134,6 @@ export class StaticWebsite extends AwsConstruct {
                 viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
                 functionAssociations: functionAssociations,
             },
-            errorResponses: [this.errorResponse()],
             // Enable http2 transfer for better performances
             httpVersion: HttpVersion.HTTP2,
             certificate: certificate,
@@ -278,10 +279,10 @@ export class StaticWebsite extends AwsConstruct {
         return this.provider.getStackOutput(this.distributionIdOutput);
     }
 
-    private errorResponse(): ErrorResponse {
+    private errorResponseDocument(): string {
         // Custom error page
         if (this.configuration.errorPage !== undefined) {
-            let errorPath = this.configuration.errorPage;
+            const errorPath = this.configuration.errorPage;
             if (errorPath.startsWith("./") || errorPath.startsWith("../")) {
                 throw new ServerlessError(
                     `The 'errorPage' option of the '${this.id}' static website cannot start with './' or '../'. ` +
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by t-richard** [comment]
This seems the right way forward to me 👍

And bonus, we get better support for static website hosting by supporting static site generation tools like Jekyll, Hugo, Gatsby, Nuxt/Next static target build, etc

**Comment by piolet** [comment]
no need to handle 403 errors?
I ask, more by anticipation than to raise a problem

**Comment by t-richard** [comment]
@piolet I think 403 could not happen anymore because the bucket now has public read access

See https://github.com/getlift/lift/pull/165/files#diff-5c9d8bfb4efe3d543af639b9a5524b66d81496d2e940def74fafca2294332e06R97

**Comment by piolet** [comment]
Yes, with public access maybe.

When I tested, my bucket had remained private, that's probably why I had to manage the 403 error

**Comment by piolet** [comment]
> And bonus, we get better support for static website hosting by supporting static site generation tools like Jekyll, Hugo, Gatsby, Nuxt/Next static target build, etc

ideally, this bonus could be applied to the server side website. I know I'm repeating myself, but it happens that sites generated with server and nuxt cohabit (a subdirectory of the main site for example #94 )
I know it's not the subject of this PR, but if it can trigger an action, you never know ;)

**Comment by adriencaccia** [comment]
> @piolet I think 403 could not happen anymore because the bucket now has public read access
> 
> See [#165 (files)](https://github.com/getlift/lift/pull/165/files#diff-5c9d8bfb4efe3d543af639b9a5524b66d81496d2e940def74fafca2294332e06R97)

Yes that's exact !

File: src/constructs/aws/StaticWebsite.ts [hunk 1]
```diff
@@ -278,39 +279,26 @@ export class StaticWebsite extends AwsConstruct {
...
             if (errorPath.startsWith("./") || errorPath.startsWith("../")) {
                 throw new ServerlessError(
                     `The 'errorPage' option of the '${this.id}' static website cannot start with './' or '../'. ` +
                         `(it cannot be a relative path).`,
                     "LIFT_INVALID_CONSTRUCT_CONFIGURATION"
                 );
             }
-            if (!errorPath.startsWith("/")) {
-                errorPath = `/${errorPath}`;
-            }
```
mnapoli, line 294: This bit was removed (and the corresponding test changed): this isn't necessary anymore with S3 doing the logic?

mnapoli:
Just tested, worked well, thanks! That approach makes sense, I guess there's no way to disable the public S3 URL? (to avoid public duplicated content)

> And bonus, we get better support for static website hosting by supporting static site generation tools like Jekyll, Hugo, Gatsby, Nuxt/Next static target build, etc

That's a very good point.

Also I keep reading the code change but I don't understand why the OAI isn't deployed anymore. If you have an answer to that I'm interested. In any case, it's not important since the bucket would be public now (so the OAI isn't needed).

src/constructs/aws/StaticWebsite.ts
adriencaccia: Exactly, the correct path is resolved with or without a leading `/`

piolet:
> I guess there's no way to disable the public S3 URL? (to avoid public duplicated content)

an answer exists. Strangely enough, this is what was done before :) https://stackoverflow.com/questions/27636977/how-to-disable-amazon-s3-raw-endpoint-access

adriencaccia:
> Just tested, worked well, thanks! That approach makes sense, I guess there's no way to disable the public S3 URL? (to avoid public duplicated content)

If we want to let the bucket handle its website behavior, and not CF, I'm afraid not.

> Also I keep reading the code change but I don't understand why the OAI isn't deployed anymore. If you have an answer to that I'm interested. In any case, it's not important since the bucket would be public now (so the OAI isn't needed).

That's it yes, the OAI will no longer be needed since the bucket handles its static website access.

mnapoli:
👍  thanks for all the answers!

PR #168, opened 2022-02-18
Add Server-Side encryption option for SQS, fixes #138

### LINKED ISSUE #138: SQS Server side encryption
Hey,

I want to add service side encryption to the queues created by lift, specifically the KmsMasterKeyId. Is this possible, if not can this property be exposed?

salemdar:
Extended SQS options with `encryption` and `encryptionKey` to enable Server-Side encryption. 
If `encryption` is set to `kmsManaged`, key will be managed by SQS. 
If it is set to `kms`, then `encryptionKey` will be required.

Closes #138

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ dist
 node_modules
 .serverless/
 package-lock.json
+.idea
--- a/docs/queue.md
+++ b/docs/queue.md
@@ -316,6 +316,31 @@ constructs:
         delay: 60
 ```
 
+### Encryption
+
+Turn on server-side encryption for the queue.
+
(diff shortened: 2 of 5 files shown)


mnapoli:
Thank you!

webmozart:
Is it possible to pass a key defined in my CloudFormation resources some how?

mnapoli:
@webmozart Hey! Judging by the docs & code in the PR (I'm not using that feature) I don't think this is possible. The construct seems to be creating the key itself, and seems to be expecting a key value (not an existing key ID or ARN).

webmozart:
@mnapoli Thanks!

PR #175, opened 2022-03-16
Add single-page-app construct

### LINKED ISSUE #172: [Static Website] Getting 404 on nested paths, and security headers are not applied
# Description

When trying to directly access a nested path of a SPA deployed with the static website construct, we get a 404 response and the security headers are not applied.

Although a 404 is received, the content is the page is still the desired one.

## Regarding the 404:
When building a SPA with a modern framework, we want to leave the handling of "bad routes" to the application itself. Since we cannot know in advance which routes are served by the application, I think a better approach would be to always return a 200 on nested routes.

## Why we get the 404
The construct is deploying a Cloudfront with a S3 with website hosting as an origin. The default behavior of a S3 with website hosting when trying to request a non-existing file is to return a 404 and display the error document if it is set.
For SPAs, we want this error document to be the same as the root document of our app. Lift is configured to work like this by default.
So, when hitting a nested route, Cloudfront receives a 404 from S3 and then return it to the client.

## Why the security headers are not applied
The security headers are applied thanks to a Cloudfront function deployed by Lift, which is set up to be invoked by viewer response events.
Per the [Restrictions on edge functions docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/edge-functions-restrictions.html), *CloudFront does not invoke edge functions for viewer response events when the origin returns HTTP status code 400 or higher.*
Thus when trying to get a nested route, Cloudfront will not apply the function because S3 (the origin) is returning a 404.

# Proposed solution
Adding a Lambda@Edge triggered by the origin response trigger to rewrite 4xx to 200 will solve the problem.
This is a common use case for Lambda@Edge, per [the docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-examples.html#lambda-examples-update-error-status-examples).
I will submit a PR following this issue to implement it !

Kudos to @CorentinDoue and @guillaumem-theodo for finding this bug !

# How to Reproduce

A created a single file SPA with react and nested routes to showcase the problem.
Accessing the deployed cloudfront domain with the nested routes `/login` or `/register` will result in the bug described above.

Lift config in `serverless.yml`:

```yaml
constructs:
  landing:
    type: 'static-website'
    path: 'build'
```

`build/index.html`

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8" />
    <link rel="shortcut icon" type="image/jpg" href="logo.svg" />

    <!-- Note: When deploying to production, replace "development.js"
      with "production.min.js" in each of the following tags -->

    <!-- Load React and React DOM -->
    <!-- See https://reactjs.org/docs/add-react-to-a-website.html to learn more -->
    <script
      src="https://unpkg.com/react@>=17/umd/react.development.js"
      crossorigin
    ></script>
    <script
      src="https://unpkg.com/react-dom@>=17/umd/react-dom.development.js"
      crossorigin
    ></script>

    <!-- Load history -->
    <script
      src="https://unpkg.com/history@5/umd/history.development.js"
      crossorigin
    ></script>

    <!-- Load React Router and React Router DOM -->
    <script
      src="https://unpkg.com/react-router@6/umd/react-router.development.js"
      crossorigin
    ></script>
    <script
      src="https://unpkg.com/react-router-dom@6/umd/react-router-dom.development.js"
      crossorigin
    ></script>

    <!-- Load babel standalone -->
    <script src="https://unpkg.com/@babel/standalone/babel.min.js"></script>
  </head>
  <body>
    <div id="root"></div>
    <script type="text/babel" data-type="module">
      const Link = ReactRouterDOM.Link;
      const Route = ReactRouterDOM.Route;
      const BrowserRouter = ReactRouterDOM.BrowserRouter;
      const Routes = ReactRouterDOM.Routes;

      const App = () => (
        <BrowserRouter>
          <Routes>
            <Route path="/" element={<Home />} />
            <Route path="/login" element={<Login />} />
            <Route path="/register" element={<Register />} />
          </Routes>
          <ul>
            <li>
              <Link to="/">Home</Link>
            </li>
            <li>
              <Link to="/login">Login</Link>
            </li>
            <li>
              <Link to="/register">Register</Link>
            </li>
          </ul>
        </BrowserRouter>
      );

      const Home = () => <h1>Home</h1>;
      const Login = () => <h1>Login</h1>;
      const Register = () => <h1>Register</h1>;

      ReactDOM.render(<App />, document.querySelector("#root"));
    </script>
  </body>
</html>
```

### Additional Information

Versions of npm packages:

```
  npmPackages:
    serverless: ^3.2.0 => 3.2.0 
    serverless-lift: ^1.12.1 => 1.12.1 
```

adriencaccia:
Fix #172


mnapoli:
Thank you for opening this! I'll review this tomorrow!

Just a question because I got lost with #172: is this PR using Lambda@Edge? From a quick search it doesn't seem so.

adriencaccia:
> Just a question because I got lost with https://github.com/getlift/lift/issues/172: is this PR using Lambda@Edge? From a quick search it doesn't seem so.

No :smiley: !

I managed to make the SPA redirection work by only adding a cloudfront function triggered by the viewer request event.

mnapoli:
Amazing!

src/constructs/aws/SinglePageApp.ts
```diff
@@ -0,0 +1,57 @@
...
+            { EventType: FunctionEventType.VIEWER_REQUEST, FunctionARN: viewerRequestFunction.functionArn },
+        ]);
+    }
+
+    public createViewerRequestCloudfrontFunction(): cloudfront.Function {
+        /**
+         * CloudFront function that redirects nested paths to /index.html and
+         * let static files pass.
+         */
+        const code = `var REDIRECT_REGEX = /^[^.]+$|\\.(?!(css|gif|ico|jpg|js|png|txt|svg|woff|woff2|ttf|map|json)$)([^.]+$)/;
```
mnapoli, line 41: Just to be 100% clear, we'll need to update this to whitelist new extensions that we may forget? (e.g. `jpeg` is not in there, I'm sure there are plenty we might not think about right?)

src/constructs/aws/SinglePageApp.ts
```diff
@@ -0,0 +1,57 @@
...
+
+function handler(event) {
+    var uri = event.request.uri;
+    var isUriToRedirect = REDIRECT_REGEX.test(uri);
+
+    if (isUriToRedirect) event.request.uri = "/index.html";
```
mnapoli, line 47: Again, just to be 100% explicit: scenarios where multiple root files aren't supported, right? (e.g. `/index.html`, and `/demo/index.html` for a separate "demo" website under the same domain)

src/constructs/aws/StaticWebsite.ts
```diff
@@ -1,333 +1,51 @@
...
-        security: {
-            type: "object",
-            properties: {
-                allowIframe: { type: "boolean" },
-            },
-            additionalProperties: false,
-        },
-        errorPage: { type: "string" },
+        ...COMMON_STATIC_WEBSITE_DEFINITION.properties,
         redirectToMainDomain: { type: "boolean" },
```
mnapoli, line 14: Any reason why this property isn't on the new construct as well?

src/constructs/aws/abstracts/StaticWebsiteAbstract.ts
```diff
@@ -0,0 +1,316 @@
...
+         * CloudFront function that manipulates the HTTP responses to add security headers.
+         */
+        const code = `function handler(event) {
+    var response = event.response;
+    response.headers = Object.assign({}, ${jsonHeaders}, response.headers);
+    return response;
+}`;
+
+        return new cloudfront.Function(this, "ViewerResponseFunction", {
+            functionName: `${this.provider.stackName}-${this.provider.region}-${this.id}-viewerResponse`,
```
mnapoli, line 312: Small nitpick but can we keep this one named `ResponseFunction`? That will minimize diffs/changes in tests, and in CloudFront Functions there is not "Origin" response function so the distinction is not needed. So anything that lets us have simpler naming/concepts is welcome.

```suggestion
        return new cloudfront.Function(this, "ResponseFunction", {
            functionName: `${this.provider.stackName}-${this.provider.region}-${this.id}-response`,
```

src/constructs/aws/SinglePageApp.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    return event.request;
+}`;
+
+        return new cloudfront.Function(this, "ViewerRequestFunction", {
+            functionName: `${this.provider.stackName}-${this.provider.region}-${this.id}-viewerRequest`,
```
mnapoli, line 53: Same here, if we can simplify and minimize diffs that's great:

```suggestion
        return new cloudfront.Function(this, "RequestFunction", {
            functionName: `${this.provider.stackName}-${this.provider.region}-${this.id}-request`,
```

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
...
+            "request": "launch",
+            "console": "integratedTerminal",
+            "internalConsoleOptions": "neverOpen",
+            "disableOptimisticBPs": true,
+            "cwd": "${workspaceFolder}",
+            "runtimeExecutable": "yarn",
+            "args": ["jest", "--runInBand", "--watchAll=false"]
+        }
+    ]
+}
```
mnapoli, line 15: Is this a shared config? Or accidental commit?

docs/single-page-app.md
```diff
@@ -0,0 +1,230 @@
...
+```yaml
+constructs:
+    landing:
+        # ...
+        errorPage: error.html # can be any HTML file in your project
+```
+
+When a browser requests the URL of a non-existing file, the `error.html` file will be served with a 404 response status.
+
+Do not use this setting when doing JavaScript URL routing: this will break URL routing.
```
mnapoli, line 212: Should we remove the `errorPage` option from the single-page app construct? Any reason to keep it?

src/constructs/aws/SinglePageApp.ts
adriencaccia: Yes that's it.
I have taken the list from https://docs.aws.amazon.com/amplify/latest/userguide/redirects.html#redirects-for-single-page-web-apps-spa, as recommended by @fredericbarthelet in his [comment](https://github.com/getlift/lift/issues/172#issuecomment-1055175111).
In my opinion, we can trust amplify for the relevance of the list, wdyt ? (I'll add the `jpeg` though nice catch)

mnapoli: Thanks, could you add a link to that URL in the comment (in the code)? That way we have a reference to check for updates. I think this makes a lot of sense to follow the AWS behavior 👍

src/constructs/aws/SinglePageApp.ts
adriencaccia: Yes, that's right. All paths not representing a static file will be solely redirected to `/index.html`

src/constructs/aws/StaticWebsite.ts
adriencaccia: Since the implementation of the feature enabled `redirectToMainDomain` is already done in a viewer request function, I removed it from the SPA construct to simplify the code. I also thought that this feature was not needed for a SPA.
I can modify the code to add it if you think it is still useful :wink:

src/constructs/aws/abstracts/StaticWebsiteAbstract.ts
adriencaccia: Sure no problem !

The reason I made this refactoring in the first place is because I started implemeting a solution with a Lamda@Edge triggered by the origin response trigger, thus I needed the function names to be really explicit.

.vscode/launch.json
adriencaccia: This is a shared config, allowing Lift devs using VSCode to have a strong jest integration:
image

docs/single-page-app.md
adriencaccia: There are no reason to keep it. I will remove it :wink:

src/constructs/aws/SinglePageApp.ts
adriencaccia: Done

docs/single-page-app.md
adriencaccia: Done

src/constructs/aws/StaticWebsite.ts
mnapoli: Oh I would imagine it's useful for SPAs too? That feature is IMO kind of always needed, because websites can be typed in browsers by humans with `www.` and without (because there's no real standard, humans are used to both formats).

So I would expect almost all websites to redirect to their main domain (either with or without `www`)?

.vscode/launch.json
mnapoli: Got it thanks!

mnapoli (commented):
Thanks for addressing all this so quickly!

cc @fredericbarthelet if you want to review

I'll try out the component early next week (just to be sure) but except that I'm all good here!

docs/single-page-app.md
```diff
@@ -0,0 +1,213 @@
...
+![](img/static-website.png)
+
+Additionally, every time `serverless deploy` runs, Lift:
+
+- uploads all files from the configured directory to the S3 bucket
+- invalidates the CloudFront cache so that the new version of the website is live
+
+To learn more about the architecture of this construct, [read this article](https://medium.com/serverless-transformation/static-websites-on-aws-designing-lift-1db94574ba3b).
+
+_Note: the S3 bucket is public and entirely managed by Lift. Do not store or upload files to the bucket, they will be removed by Lift on the next deployment. Instead, create a separate bucket to store any extra file._
```
fredericbarthelet, line 50: Do you still require the S3 bucket to be public with the Cloudfront function construct rebuilding URL to match `index.html` ? This is a security concern for some companies to not allow S3 content to be accessible bypassing Cloudfront alltogether

docs/single-page-app.md
```diff
@@ -0,0 +1,213 @@
...
+    landing:
+        # ...
+        domain:
+            - mywebsite.com
+            - app.mywebsite.com
+```
+
+### Allow iframes
+
+By default, as recommended [for security reasons](https://scotthelme.co.uk/hardening-your-http-response-headers/#x-frame-options), the static website cannot be embedded in an iframe.
```
fredericbarthelet, line 199:
```suggestion
By default, as recommended [for security reasons](https://scotthelme.co.uk/hardening-your-http-response-headers/#x-frame-options), the single page app cannot be embedded in an iframe.
```

docs/static-website.md
```diff
@@ -2,7 +2,6 @@
 
 The `static-website` construct deploys:
```
fredericbarthelet, line 3:
```suggestion
The `static-website` construct deploys **plain static websites** composed of HTML files and assets (CSS, JS…)
```

No need for a list anymore :)

src/constructs/aws/abstracts/StaticWebsiteAbstract.ts
```diff
@@ -0,0 +1,316 @@
...
+            throw new ServerlessError(
+                `Invalid configuration for the static website '${id}': if a domain is configured, then a certificate ARN must be configured in the 'certificate' option.\n` +
+                    "See https://github.com/getlift/lift/blob/master/docs/static-website.md#custom-domain",
+                "LIFT_INVALID_CONSTRUCT_CONFIGURATION"
+            );
+        }
+
+        const bucket = new Bucket(this, "Bucket", {
+            // Enable static website hosting
+            websiteIndexDocument: "index.html",
```
fredericbarthelet, line 90: The commit https://github.com/getlift/lift/commit/e5da8c9daa658dd33395961202fe7900a0bfcaad should be reverted as it is not used anymore for SPA to work properly

fredericbarthelet:
@adriencaccia thanks for putting this PR together :)

LGTM, except a few minor comments and https://github.com/getlift/lift/commit/e5da8c9daa658dd33395961202fe7900a0bfcaad that can then be reverted to reduce boilerplate and avoid confusion on which mechanism is in place to redirect to index (in the PR current state, it looks like both Cloudfront and S3 are in charge of the redirect)

fredericbarthelet (changes requested):
Can you add https://github.com/getlift/lift/commit/e5da8c9daa658dd33395961202fe7900a0bfcaad revert to this PR ?

docs/single-page-app.md
mnapoli: If we can skip making the bucket public that would be a huge win indeed.

fredericbarthelet:
@adriencaccia as seen together, reverting back to OAI and private bucket (removing website hosting on S3 bucket) will work as intended with your new Cloudfront function redirecting SPA path to `/`. This fix + enabling `redirectToMainDomain` on the new construct and this PR will be good to go :)

Thanks again for your contribution, let me know if @mnapoli and I can help in any way.

mnapoli:
> @adriencaccia as seen together, reverting back to OAI and private bucket (removing website hosting on S3 bucket) will work as intended with your new Cloudfront function redirecting SPA path to `/`. This fix + enabling `redirectToMainDomain` on the new construct and this PR will be good to go :)
> 
> Thanks again for your contribution, let me know if @mnapoli and I can help in any way.

+1, yes I'd say OAI has a great advantage (private bucket). If we can get these things ☝️  IMO we're good to go!

src/constructs/aws/StaticWebsite.ts
adriencaccia: Done !

docs/single-page-app.md
adriencaccia: Done !

adriencaccia:
Hey @mnapoli @fredericbarthelet !

I made all the changes we talked about!
The origin bucket is now private. The main change is that users will get a 403 when trying to access a non existing file (I am writing this here so that we are all on the same page).

The construct is working properly on my end, can you test it ?

fredericbarthelet:
@adriencaccia LGTM :) !
Thanks for your contribution. I'll merge and release a new version of Lift right away.

fredericbarthelet (approved):
@adriencaccia LGTM :) !
Thanks for your contribution. I'll merge and release a new version of Lift right away.

mnapoli:
🎉  thanks @adriencaccia!

PR #177, opened 2022-04-04
Add options for visibility timeout and content based deduplication

andremonteiro95:
We have run into some use cases where we need to manually set the visibility timeout and turn off content-based duplication, as such, we had the need to configure these directly in the serverless.yml.

### DIFF
--- a/src/constructs/aws/Queue.ts
+++ b/src/constructs/aws/Queue.ts
@@ -47,7 +47,13 @@ const QUEUE_DEFINITION = {
             minimum: 0,
             maximum: 300,
         },
+        visibilityTimeout: {
+            type: "number",
+            minimum: 0,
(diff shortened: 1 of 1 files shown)


andremonteiro95:
Hi @fredericbarthelet, thanks for your reply.

I guess the first proposal is more in line with what Serverless already supports (`resources.extensions`). 

Can you give me an example of how it would work with an SQS resource? I looked at the [AWS Cloudformation Resource Reference](https://www.serverless.com/framework/docs/providers/aws/guide/resources/#aws-cloudformation-resource-reference) table and I didn't find a reference for SQS, I don't know if you are relying on these.

seantcanavan:
@fredericbarthelet is there any documentation on how to disable content based deduplication with #191?

PR #179, opened 2022-04-04
Add WebAcl construct support for cloudfront

### LINKED ISSUE #178: Add support for WebACL (WAF2 and WAF Classic)
It's sometimes necessary to protect our cloudfront distribution with a
WAF2/WAF Classic web ACL. A very first simple approach would be to allow linking
an existing WAF resource.

I propose something like:

```yaml
constructs:
    website:
        # ...
        webAclId: arn:aws:wafv2:us-east-1:{account}:global/webacl/{webacl_name}/{webacl_id}
```

tonivdv:
fixes #178

It's sometimes necessary to protect our cloudfront distribution with a
WAF2/WAF Classic web ACL. A very first simple approach would be to allow linking
an existing WAF resource.

Basically following should be possible:

```yaml
constructs:
    website:
        # ...
        webAclId: arn:aws:wafv2:us-east-1:{account}:global/webacl/{webacl_name}/{webacl_id}
```

## Help

I never contributed on any serverless plugin, so could someone help me out for adding tests and how I can run them on my local machine as out of the box the tests fail. And in the contributing guidelines I didn't find anything specific to that. Thx

- [ ] Adding tests

### DIFF
--- a/docs/static-website.md
+++ b/docs/static-website.md
@@ -192,6 +192,19 @@ constructs:
 
 The first domain in the list will be considered the main domain. In this case, `mywebsite.com` will redirect to `www.mywebsite.com`.
 
+### Web ACL
+
+If you have a Web ACL configured in WAF2 or WAF Classing you can link it to the cloudfront distribution:
+
+```yaml
+constructs:
+    website:
+        # ...
+        webAclId: arn:aws:wafv2:us-east-1:{account}:global/webacl/{webacl_name}/{webacl_id}
+```
+
+NOTE: For WAF2, you need to specify the arn as demonstrated in the example. For WAF Classic you must set the ID. 
+
 ### Error page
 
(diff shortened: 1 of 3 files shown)


tonivdv:
@fredericbarthelet could you approve the workflow ? I would be able then to use the test execution from the Github CI as I'm not able to run them locally. Thx

mnapoli:
We discussed this PR earlier with @fredericbarthelet and we came up with the idea illustred in #191.

Because we don't want every construct to turn into something with hundreds of options, but also we don't want to limit users. WDYT?

tonivdv:
@mnapoli using resources extensions as stated in #191 is more than fine. This PR can be closed then and we can then refer issue #178 to #191

fredericbarthelet:
Discussed in #191

PR #180, opened 2022-04-07
Add ability to define the distribution comment

j0k3r:
I wanted to be able to define the comment of the distribution to follow our internal naming.
So I've added a `comment` options that will be taken as the distribution comment if specified.

I added that option for:
- static-website
- server-side-website

I wasn't 100% sure about the name `comment`. I'm open to new name :)

### DIFF
--- a/docs/server-side-website.md
+++ b/docs/server-side-website.md
@@ -346,6 +346,43 @@ constructs:
 
 CloudFront accepts maximum 10 headers.
 
+### Define the Distribution comment
+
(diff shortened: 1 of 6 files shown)


mnapoli:
Same as many other PRs, we'd rather avoid adding an option for every single CloudFormation property. So we're thinking about going with #191. WDYT?

j0k3r:
It'll _solve_ the problem, yep!
I'm closing my PR then.

PR #193, opened 2022-04-27
Update cdk to v2

fredericbarthelet:
This PR upgrades CDK from v1 to v2.
This considerably reduces the quantity of library in Lift dependencies. It also resolves L1 version conflict among various packages.

[CDK migration guide](https://docs.aws.amazon.com/cdk/v2/guide/migrating-v2.html) was used to perform this upgrade.

The resulting dependencies includes :

- `aws-cdk-lib` for all constructs libraries
- `constructs` to be able to reference `Construct` interface
- `@aws-cdk/aws-apigatewayv2-alpha` required for the `Webhook` construct in order to provision an `HttpApi`, which was already in alpha in CDK v1, and remains in alpha in CDK v2

A few tests were adjusted to accommodate this update :

- `securityGroupName` property on `SecurityGroup` construct [was deprecated in favor of `securityGroupId`](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-ec2.SecurityGroup.html#securitygroupnamespan-classapi-icon-api-icon-deprecated-titlethis-api-element-is-deprecated-its-use-is-not-recommended%EF%B8%8Fspan). This affects the generated Cloudformation intrinsic function used to reference such resource.
- CloudFront minimum TLS requirements were upgraded from `TLSv1.2_2019` to `TLSv1.2_2021`

In addition, the new bootstrap version check enabled by CDK v2 was disabled implicitly following [this guideline](https://github.com/aws/aws-cdk/issues/17942#issuecomment-992295898)

AWS CDK v2 also removes support for node 10 and 12. Node 10 support was already removed from Lift, however this change will prevent users currently using Lift with node 12 to upgrade.


t-richard (approved):

fredericbarthelet:
Thanks @mnapoli and @t-richard for your reviews. i however had to add an additional commit afterwards removing support for Node 12, as such node version is not supported by CDK v2. Do you agree with dropping support for Node 12 (end of life for Node 12 is due in 3 days -> https://nodejs.org/en/about/releases/)

mnapoli:
👍  sounds good!

t-richard:
Seems fine to me too. Node 12 is EOL and most projects using Lift are probably recent projects using Node 14 already.

fredericbarthelet:
Great, thanks for the feedbacks :). Merging and releasing with a changelog mentioning the summary of the BC above.

PR #198, opened 2022-05-02
Allow more proxy related headers to pass through

yangm97:
Some applications such as wordpress get caught in a redirect loop when these headers are not present.

### DIFF
--- a/src/constructs/aws/ServerSideWebsite.ts
+++ b/src/constructs/aws/ServerSideWebsite.ts
@@ -389,6 +389,9 @@ export class ServerSideWebsite extends AwsConstruct {
             "Referer",
             "User-Agent",
             "X-Requested-With",
+            "X-Forwarded-For",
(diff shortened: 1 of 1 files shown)


t-richard:
Cloudfront is limited to 10 headers unfortunately. As far as i'm concerned, I never faced an issue because of this. 

Could you elaborate on the issue you are facing? Have you tested this fix in your environment?

mnapoli:
The thing is that these headers are not really interesting (API Gateway is always on HTTPS) and AFAIK always populated by API Gateway anyway? (or maybe I'm mistaking it with something else)

t-richard:
> AFAIK always populated by API Gateway anyway

This is what I was thinking too

yangm97:
> AFAIK always populated by API Gateway anyway? 

You're correct. I guess they added with v2? I vaguely remember having to workaround something like this in the past...

Sorry for the noise.

PR #199, opened 2022-05-03
feat(single-page-app): add pdf to redirect regex

GuillaumeLagrange:
When trying to serve a pdf as a static file with the single page app contruct, the cloudfront distribution would redirect to index.html because the pdf file extension was not whitelisted. 

This PR adds the pdf file extension to the whitelist so that users can serve pdf files statically.


GuillaumeLagrange:
Indeed, should have caught that locally :+1: 

Fixed locally, pipeline should be good now

fredericbarthelet (approved):

fredericbarthelet:
LGTM, thanks @GuillaumeLagrange

PR #200, opened 2022-05-03
Implement constructs extensions

### LINKED ISSUE #191: Implement resources extensions to allow further customization of Lift Constructs underlying CFN resources
### Use-case

They are a lot of issue regarding the addition of a new property on existing Lift constructs to hydrate a specific underlying CFN property. All issues are tagged with `custom property` label. This issue aims at finding a better way to extend constructs.

Current recommendation to do just that leverages [`resources.extensions` property](https://www.serverless.com/framework/docs/providers/aws/guide/resources/#override-aws-cloudformation-resource) to override specific properties. However, this method has the disadvantage to rely on LogicalIDs generated by the CDK in Lift:

```yml
# serverless.yml
service: my-app
provider:
    name: aws

constructs:
    landing:
        type: static-website
        path: public

plugins:
    - serverless-lift 
    
resources:
    extensions:
        LandingBucketF057328A:
            Properties:
                ObjectLockEnabled: true
```

This issue aims at finding a way not to have to version LogicalIDs of resources generated by Lift.

### Implementations proposals

There are multiple possible ways to declare extensions

**Using a global `lift.extensions` property**
```yml
# serverless.yml
service: my-app
provider:
    name: aws

constructs:
    landing:
        type: static-website
        path: public

plugins:
    - serverless-lift 
    
lift:
    extensions:
        landing.bucket:
            Properties:
                ObjectLockEnabled: true
```

At deployment time, Lift will replace `landing.bucket` with the S3 bucket LogicalId generated by Lift for the Static Website construct

**Using a construct scoped `extensions` property**
```yml
# serverless.yml
service: my-app
provider:
    name: aws

constructs:
    landing:
        type: static-website
        path: public
        extensions:
            bucket:
                Properties:
                    ObjectLockEnabled: true

plugins:
    - serverless-lift 
```

For both exemples, each construct will expose a set of resources corresponding to final CFN resources created.
For exemple, in the storage construct, `bucket` property will reference underlying `[AWS::S3::Bucket](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket.html) resource`

fredericbarthelet:
Fixes #191


t-richard:
I love the idea!

I don't have a clear idea of what the ideal solution would be but here are some thoughts to get the discussion started around this.

Do we want to expose Cloudformation properties? Or would it be possible to expose underlying CDK properties? I guess the latter is harder because not everything is a property, some things need to be created as separate objects or as method calls.

From a Lift developer perspective, wouldn't it be more friendly to have a generic way to expose and merge config built in the provider instead of each individual construct?

Would this be as `resources.extensions` in Serverless framework? I now that in sls, the extensions are only replacing keys at the first level. This would be problematic for Lift as some Cloudformation resources are poorly designed on AWS side like Cloudfront distributions that have a single `DistributionConfig` key in properties. 
Overriding the static website or server-side website would get way harder as you would need to redifine every property to make it work with undermines the benefits of using Lift.

fredericbarthelet:
> Would this be as `resources.extensions` in Serverless framework? I now that in sls, the extensions are only replacing keys at the first level. This would be problematic for Lift as some Cloudformation resources are poorly designed on AWS side like Cloudfront distributions that have a single `DistributionConfig` key in properties. Overriding the static website or server-side website would get way harder as you would need to redifine every property to make it work with undermines the benefits of using Lift.

Thanks for your feedback @t-richard :) I made Lift's API to provide extensions entry point as simple to enrich as possible. I now completed first version of this PR for you and @mnapoli to have a deeper look on the internals. Additional entry point can easily be added for the extensions as use cases pops up.

For exemple, if you want to provide for Lift users at later stage an option to extend the `OriginRequestPolicy` of the underlying CloudFormation resource within `server-side-website` construct, you'll "just" have to add the following line in the `extend` method of the construct :

```ts
    extend(): Record<string, CfnResource> {
        return {
            distribution: this.distribution.node.defaultChild as CfnDistribution,
            originRequestPolicy: this.originRequestPolicy.node.defaultChild a CfnCachePolicy
        };
    }
```

More complex selector that the defaultChild of each CDK node can be used to reference the underlying in depth resources of the CloudFormation template. Would welcome any PR in a second stage implementing those to ensure no extensions requires a whole bunch of Lift best practice to be reimplemented manually ;)

mnapoli (approved):
That looks awesome, and the docs too!

Thanks 🎉  @fredericbarthelet

docs/server-side-website.md
```diff
@@ -346,6 +346,30 @@ constructs:
...
+            distribution:
+                Properties:
+                    Comment: Landing distribution
+```
+
+### Available extensions
+
+| Extension key 	| CloudFormation resource       	| CloudFormation documentation                                                                                           	|
+|---------------	|-------------------------------	|------------------------------------------------------------------------------------------------------------------------	|
+| distribution  	| AWS::CloudFront::Distribution 	| [Link](https://docs.aws.amazon.com/fr_fr/AWSCloudFormation/latest/UserGuide/aws-resource-cloudfront-distribution.html) 	|
```
piolet, line 370: hello,

may I ask a question, more than a comment?
For this type of builder, we will also create an S3 bucket, will it be possible to adjust it, like the "storage" builder, or will we only be able to adjust the distribution?

fredericbarthelet: Absolutely possible :) I'll merge this one, if you want to provide a PR on the basis of this new code base to implement the storage extension on this construct, you're welcome to :)

piolet: I would love to, but I can't compile the project properly, nor can I run the unit tests without errors on my side (on the branch master). So it's hard to try implementations :/

Issue #192, opened 2022-04-25
Static websites fail to rewrite URL for sub directories


tonivdv:
### Description

I think that since the changes from #175 there is BC break introduced for the 'static-website' type deploy. When deploy a static website that has an `index.html` in a sub directory , that file will not be loaded automatically anymore. To view to page you have to explicitly provide the index.html in the url.

❌  - https://some-website/sub-directory
✅  - https://some-website/sub-directory/index.html

From what I could see, is that previously the static-website was deployed with the [S3 static website feature](https://docs.aws.amazon.com/AmazonS3/latest/userguide/WebsiteHosting.html). Now it doesn't causing it to fail.

### How to Reproduce

Deploy a static website with some sub directories and an index.html page in it and deploy like a static website

```yaml
constructs:
  app:
    type: static-website
```

If you then open the url with the subdirectory you will get following error:

```xml
<Error>
<Code>AccessDenied</Code>
<Message>Access Denied</Message>
<RequestId>XXXX</RequestId>
<HostId>XXXX</HostId>
</Error>
```

### Additional Information

Additionally, the problem is well explained [here](https://github.com/aws-samples/amazon-cloudfront-functions/tree/main/url-rewrite-single-page-apps) and with a possible solution.

tonivdv:
> I still think /directory/ should work though, can you confirm if it doesn't work?

You are correct, the `/directory/` does work. However `/directory` does not. 

> That being said, I'm not against switching static-website back again to S3 website hosting -> since it was the case in 1.12.1 I'm guessing there will be no downside except https://github.com/getlift/lift/issues/172, and https://github.com/getlift/lift/issues/172 was fixed by introducing the single-page-app construct (this construct will not be impacted).

I played around a bit and there is something very strange that I can't explain at the moment. When I deploy my app with version 1.12.1, I can see my web page when calling an url with `/directory` ... 

However, when I deploy with 1.14.0, and I manually reconfigure everything exactly the way it is done with 1.12.1, it still refuses to show the page with `/directory`.

So I really don't understand what is still different underlying.

PR #202, opened 2022-05-06
fix(static-website): re-enable static website hosting

fargito:
The introduction of the single page app construct introduced a breaking change in the static website construct by removing the static website hosting.

This PR re-enables static website hosting and public read access for static buckets.

I also edited the static website tests in order to make them more strict.

Fixes https://github.com/getlift/lift/issues/192

Also see: 
- the initial issue: https://github.com/getlift/lift/issues/172
- the related PR: https://github.com/getlift/lift/pull/175

@adriencaccia

### DIFF
--- a/src/constructs/aws/StaticWebsite.ts
+++ b/src/constructs/aws/StaticWebsite.ts
@@ -57,11 +57,6 @@ export class StaticWebsite extends StaticWebsiteAbstract {
             code: cloudfront.FunctionCode.fromInline(code),
         });
     }
-    /**
-     * Overrides the default `getBucketProps` from the abstract class
-     *
-     * @returns bucketProps
-     */
 
     getBucketProps(): BucketProps {
         return {
--- a/src/constructs/aws/StaticWebsite.ts
+++ b/src/constructs/aws/StaticWebsite.ts
@@ -2,6 +2,8 @@ import * as cloudfront from "aws-cdk-lib/aws-cloudfront";
 import { FunctionEventType } from "aws-cdk-lib/aws-cloudfront";
 import type { Construct as CdkConstruct } from "constructs";
 import type { AwsProvider } from "@lift/providers";
+import type { BucketProps } from "aws-cdk-lib/aws-s3";
+import { RemovalPolicy } from "aws-cdk-lib";
 import { redirectToMainDomain } from "../../classes/cloudfrontFunctions";
 import { getCfnFunctionAssociations } from "../../utils/getDefaultCfnFunctionAssociations";
 import type { CommonStaticWebsiteConfiguration } from "./abstracts/StaticWebsiteAbstract";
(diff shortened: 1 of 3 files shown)


src/constructs/aws/StaticWebsite.ts
```diff
@@ -17,7 +19,18 @@ export class StaticWebsite extends StaticWebsiteAbstract {
...
+            // Enable static website hosting
+            websiteIndexDocument: "index.html",
+            websiteErrorDocument: configuration.errorPage,
+            // public read access is required when enabling static website hosting
+            publicReadAccess: true,
+            // For a static website, the content is code that should be versioned elsewhere
+            removalPolicy: RemovalPolicy.DESTROY,
+        };
+
+        super(scope, id, configuration, provider, bucketProps);
```
fredericbarthelet, line 33: `super` should always be called first in any class implementing an abstract class in order to avoid anti-pattern in composition. If there are 2 different bucket configuration for SPA and static website, bucket creation should be the scope of those specific construct. Cloudfront configuration common resources should remain within the abstract `StaticWebsiteAbstract` construct.

src/constructs/aws/StaticWebsite.ts
```diff
@@ -17,7 +19,18 @@ export class StaticWebsite extends StaticWebsiteAbstract {
         protected readonly configuration: CommonStaticWebsiteConfiguration,
         protected readonly provider: AwsProvider
     ) {
-        super(scope, id, configuration, provider);
+        // we need to build specific bucket props to pass to the abstract static website constructor
+        const bucketProps: BucketProps = {
+            // Enable static website hosting
+            websiteIndexDocument: "index.html",
+            websiteErrorDocument: configuration.errorPage,
```
fredericbarthelet, line 26: isn't this duplicate with https://github.com/getlift/lift/blob/f715c8691a622092b50760c37c53344b3aa325b8/src/constructs/aws/abstracts/StaticWebsiteAbstract.ts#L269 ?

src/constructs/aws/StaticWebsite.ts
```diff
@@ -17,7 +19,18 @@ export class StaticWebsite extends StaticWebsiteAbstract {
...
+            // public read access is required when enabling static website hosting
+            publicReadAccess: true,
```
fredericbarthelet, line 28: as stated in https://github.com/getlift/lift/issues/192#issuecomment-1111206102, it was actually an improvement not to have public access on the assets within S3. Is there another solution ? Like using CF function instead ?

fargito: It would be an improvement indeed but I couldn't make it work :/
Also since a static website has no authentication, it is not dramatic to allow public read access. See https://docs.aws.amazon.com/AmazonS3/latest/userguide/WebsiteAccessPermissionsReqd.html

mnapoli (approved):
Thanks a lot for working on this!

All good for me, I'll let @fredericbarthelet review the changes.

src/constructs/aws/StaticWebsite.ts
```diff
@@ -55,4 +57,21 @@ export class StaticWebsite extends StaticWebsiteAbstract {
             code: cloudfront.FunctionCode.fromInline(code),
         });
     }
+    /**
+     * Overrides the default `getBucketProps` from the abstract class
+     *
+     * @returns bucketProps
```
fredericbarthelet, line 63: We're not using JS Doc to provide return type of function, only return Type from Typescript. Can you remove this line ?

fredericbarthelet (approved):

PR #206, opened 2022-05-17
Add CSP header

fredericbarthelet:
Following recommendations from [AWS Documentation](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/example-function-add-security-headers.html), header [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) should be added in CloudFront responses.


mnapoli:
What will be the actual impact of setting that new header?

The way I naively interpret it is that:

- styles (CSS), scripts (JS) and images cannot be loaded from different domains
- anything else (like fonts, iframes or whatever) cannot be loaded at all

This is probably wrong right?

PR #213, opened 2022-05-27
feat(static-website): allow empty params

fargito:
Hello!

When using the `static-website` construct, I wanted to use Serverless V3 Stage params to define the settings for my static websites.

I didn't want to define a certificate and custom domains for my dev environment, so I tried to pass an empty string and an empty array instead. The Serverless framework remove `null` arguments so this wasn't an option.

Check out the PR in which I wanted to use this feature: https://github.com/swarmion/swarmion/pull/134

However these options were not handled by Lift before passing them to the CDK.

This PR adds the possibility to pass empty arguments for `certificate` and `domain` configurations.

What do you think?

### DIFF
--- a/src/constructs/aws/abstracts/StaticWebsiteAbstract.ts
+++ b/src/constructs/aws/abstracts/StaticWebsiteAbstract.ts
@@ -94,9 +94,14 @@ export abstract class StaticWebsiteAbstract extends AwsConstruct {
         const bucket = new Bucket(this, "Bucket", bucketProps);
 
         // Cast the domains to an array
-        this.domains = configuration.domain !== undefined ? flatten([configuration.domain]) : undefined;
+        // if configuration.domain is an empty array or an empty string, ignore it
+        this.domains =
+            configuration.domain !== undefined && configuration.domain.length > 0
+                ? flatten([configuration.domain])
+                : undefined;
+        // if configuration.certificate is an empty string, ignore it
         const certificate =
-            configuration.certificate !== undefined
+            configuration.certificate !== undefined && configuration.certificate !== ""
                 ? acm.Certificate.fromCertificateArn(this, "Certificate", configuration.certificate)
                 : undefined;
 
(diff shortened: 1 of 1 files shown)


mnapoli:
> I tried to pass an empty string and an empty array instead. The Serverless framework remove null arguments so this wasn't an option.

I'm not sure I understand how this was a problem? If the Framework removes the null properties that should be exactly what you want, no?

fargito:
My explanation wasn't clear, sorry.

What I need to do is parametrize some of my config:
```
  constructs: {
    app: {
      type: 'static-website',
      path: '../build',
      errorPage: '404.html',
      domain: '${param:domain}',
      redirectToMainDomain: '${param:redirectToMainDomain}',
      certificate: '${param:certificate}',
    },
  },
```

However if I try to setup my params with null values, they get cleaned out from stage params.
```
  params: {
    dev: {
      domain: [],
      redirectToMainDomain: false,
      certificate: null, // issue here
    },
    production: {
      domain: ['www.mywebsite.com', 'mywebsite.com'],
      redirectToMainDomain: true,
      certificate: 'my-certificate-arn',
    },
  },
```

If I do this, I get the following error:
image

And if I try to put the params like:
```
  params: {
    dev: {
      domain: [],
      redirectToMainDomain: false,
      certificate: '' // with an empty string
    },
    production: {
      domain: ['www.mywebsite.com', 'mywebsite.com'],
      redirectToMainDomain: true,
      certificate: 'my-certificate-arn',
    },
  },
```
I get the following error:
image

Maybe the issue is rather on the side of the Serverless Framework, what do you think?

mnapoli:
Ohhh right, yes, that's a very annoying bug: https://github.com/serverless/serverless/issues/10492

Then I'm 👍  on merging. We'll probably wait on #215 I guess?

fargito:
Yes I think it could be more convenient to wait, I'll try and update https://github.com/getlift/lift/pull/215 at the beginning of next week. Is that ok for you?

mnapoli:
Sounds perfect! Thank you!

fargito:
Waiting for #200 to be merged

PR #215, opened 2022-05-27
fix(aws): bootstrap functions key in services and not configurationInput

fargito:
All tests are currently failing on the master branch.

image

I believe the reason is that `Object.assign` to replace an empty functions config was applied to `configurationInput` and not `services`

This PR fixes this build issue.

What do you think?

### DIFF
--- a/src/providers/AwsProvider.ts
+++ b/src/providers/AwsProvider.ts
@@ -15,7 +15,12 @@ import {
     Webhook,
 } from "@lift/constructs/aws";
 import { getStackOutput } from "../CloudFormation";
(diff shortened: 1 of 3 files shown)


fargito:
Unit tests still fail but lint and type are ok...
Making progress here 😆

fredericbarthelet:
This isssue is fixed by #200 ;)

fargito:
@fredericbarthelet do you know when #200 will be merged ? This is blocking for #213

mnapoli:
FYI I just reviewed #200, looks good to me!

fargito:
Ah great! I'll rebase #213 when it is merged

PR #217, opened 2022-05-30
feat(static-website): allow empty params

fargito:
Follow-up of #213 which was unfortunately closed

What I need to do is parametrize some of my config:
```
  constructs: {
    app: {
      type: 'static-website',
      path: '../build',
      errorPage: '404.html',
      domain: '${param:domain}',
      redirectToMainDomain: '${param:redirectToMainDomain}',
      certificate: '${param:certificate}',
    },
  },
```

However if I try to setup my params with null values, they get cleaned out from stage params.
```
  params: {
    dev: {
      domain: [],
      redirectToMainDomain: false,
      certificate: null, // issue here
    },
    production: {
      domain: ['www.mywebsite.com', 'mywebsite.com'],
      redirectToMainDomain: true,
      certificate: 'my-certificate-arn',
    },
  },
```

If I do this, I get the following error:
image

And if I try to put the params like:
```
  params: {
    dev: {
      domain: [],
      redirectToMainDomain: false,
      certificate: '' // with an empty string
    },
    production: {
      domain: ['www.mywebsite.com', 'mywebsite.com'],
      redirectToMainDomain: true,
      certificate: 'my-certificate-arn',
    },
  },
```
I get the following error:
image

This PR adds the possibility to pass empty arguments for `certificate` and `domain` configurations.

### DIFF
--- a/src/constructs/aws/abstracts/StaticWebsiteAbstract.ts
+++ b/src/constructs/aws/abstracts/StaticWebsiteAbstract.ts
@@ -94,9 +94,14 @@ export abstract class StaticWebsiteAbstract extends AwsConstruct {
         const bucket = new Bucket(this, "Bucket", bucketProps);
 
         // Cast the domains to an array
-        this.domains = configuration.domain !== undefined ? flatten([configuration.domain]) : undefined;
+        // if configuration.domain is an empty array or an empty string, ignore it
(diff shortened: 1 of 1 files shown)


mnapoli (approved):
OK for me 👍

PR #218, opened 2022-05-31
Improve backend cache policy

### LINKED ISSUE #144: Cache Policy for Server-Side-Website
There is an issue with the cache policy. If there happens multiple requests per second to an url with same paths but different get-parameters, it happens that you get cache hits from cloudfront and the request result is the same, regardless of the get parameters.

Solution:
For the cache policy, all query strings have to be included into the cache key.

bytestream:
See https://github.com/getlift/lift/issues/144#issuecomment-1142111775

Untested. @mnapoli is this what you had in mind?

Closes #144

### DIFF
--- a/src/constructs/aws/ServerSideWebsite.ts
+++ b/src/constructs/aws/ServerSideWebsite.ts
@@ -2,7 +2,9 @@ import { Bucket } from "aws-cdk-lib/aws-s3";
 import type { CfnDistribution } from "aws-cdk-lib/aws-cloudfront";
 import {
     AllowedMethods,
+    CacheCookieBehavior,
     CacheHeaderBehavior,
+    CacheQueryStringBehavior,
     CachePolicy,
     Distribution,
     FunctionEventType,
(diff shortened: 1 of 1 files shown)


src/constructs/aws/ServerSideWebsite.ts
```diff
@@ -122,15 +122,6 @@ export class ServerSideWebsite extends AwsConstruct {
...
             headerBehavior: this.headersToForward(),
         });
-        const backendCachePolicy = new CachePolicy(this, "BackendCachePolicy", {
-            cachePolicyName: `${this.provider.stackName}-${id}`,
-            comment: `Cache policy for the ${id} website.`,
-            // For the backend we disable all caching by default
-            defaultTtl: Duration.seconds(0),
-            // Authorization is an exception and must be whitelisted in the Cache Policy
-            // This is the reason why we don't use the managed `CachePolicy.CACHING_DISABLED`
-            headerBehavior: CacheHeaderBehavior.allowList("Authorization"),
```
t-richard, line 132: I'm not sure about the impact on forwarding headers because there is this comment

https://github.com/getlift/lift/blob/ee5d84d3a70a879325a651ea2b3f8f497c9b804e/src/constructs/aws/ServerSideWebsite.ts#L380

mnapoli: Indeed, this isn't ideal. But related to https://github.com/getlift/lift/issues/144#issuecomment-1142111775 (and the linked Bref issue) that would solve the caching of requests CloudFront receives in the same second.

I don't see another solution.

But at the same time, I don't think `Authorization` is a header that is used a lot for server-side websites?

Also, I haven't even checked if `Authorization` is actually not forwarded when using no cache with CloudFront.

jjmontgo: I checked this myself and verified that `Authorization` is not forwarded when using the caching disabled policy.

I've had a few cases in the past when I did need the `Authorization` header.

I tested using the custom caching policy with All Cookies and All Query Strings in the cache key, and that prevented same-second requests from hitting cache.

mnapoli:
I'll be closing this PR because it unfortunately doesn't solve the problem, see https://github.com/getlift/lift/issues/144#issuecomment-1179686086 for more details, but long story short:

- we cannot switch to "caching disabled"
- setting `minTtl: 0` didn't seem to change anything in my tests
- setting `maxTtl: 1` didn't solve the problem either

Also please don't consider this as me closing the discussion. I'm closing this because I want this PR to keep the same content, because I've linked this PR in a few GitHub comments. So I'd rather we keep this one as it is (no changes to the diff), close it, and open new pull requests with alternative approaches 👍 

Let's discuss ideas in #144, and feel free to open new pull requests with any idea you may have 🙏 

Thank you for looking into it @bytestream

mnapoli:
Reopening after discussing with @fredericbarthelet, see more details here: https://github.com/getlift/lift/issues/144#issuecomment-1183210030

TL/DR: this PR is not a perfect solution, but at least an immediate improvement until we find a better solution.

Thank you @bytestream.

PR #232, opened 2022-06-22
Add support for retention period

AndrewBarba:
Add support message `retentionPeriod` when creating an SQS queue.

Use case should be self explanatory: we want to adjust the default behavior of 4 days to be shorter or longer.

### DIFF
--- a/docs/queue.md
+++ b/docs/queue.md
@@ -320,7 +320,7 @@ constructs:
 
 When a message is sent to the queue, it will be retained for up to 4 days before being automatically deleted.
 
-You can adjust this rentention period to be shorter or longer by specifing a time a seconds.
+You can adjust this rentention period to be shorter or longer by specifing a time in seconds.
 
 The maximum value is 1209600 seconds (14 days) and the minimum value is 60 seconds (1 minute).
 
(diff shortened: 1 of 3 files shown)


mnapoli:
Hey @AndrewBarba, could you expand a bit more on why you need to make it longer than 4 days?

I'm afraid this setting would impose an extra decision on users, and they would ask themselves: "do we need to store it longer than 4 days? The option's there, so there must be a reason that some users need that, what could it be?"

The original assumption in the design was that 4 days should be more than enough to _process_ messages. If the queue is backed up, that should usually be solved in the same day, or maybe the next. If your queue is backed up for more than 4 days then something serious is going on, so manual intervention (e.g. clearing the queue, or changing the retention manually) is to be expected.

Also I want to point out that we now have "extensions", so you can customize that setting today via CloudFormation directly: https://github.com/getlift/lift/blob/master/docs/queue.md#extensions We want to favor extensions for all settings we consider "advanced", i.e. that don't apply to 90% of the users.

AndrewBarba:
I don't have too many examples of longer than 4 days but this option also allows for less than 4 days of which I think there are many use cases. 

Regardless, extensions are fine and I will continue to use that. 

Also I think you're either giving your plug-in way too much credit or giving users way less credit than they deserve. Thinking that merely having a top level option is going to stop users in their tracks and overwhelm them is some wild line of thinking.

PR #252, opened 2022-08-10
watemhost

Watemlifts:
The intention is to integrate and merge different application on one platform to bridge lift control systems with a wireless  communication pot or data pot


Watemlifts:
Application possible for wireless communication and data systems of lift control hardware and software utility, upgrade and new installation

PR #273, opened 2022-10-21
Add dynamic_assets configuration

jlestel:
I use assets to upload documents that I want to make accessible via the same URL as statics and the API Gateway.

For the moment the assets are always synchronized.
Also, I want to keep uploaded files.

I added dynamic_assets to  configuration. Same structure as assets but don't try to s3sync these folders.


mnapoli:
Thank you for the PR.

That sounds like a reasonable start, but the `assets` bucket (and the whole construct) was not designed at all to store uploaded files. I think this feature might have more impact than initially thought.

For example, if there is user data in the bucket, then we don't want `retention: delete`, because we don't want to lose user data. Also, just like we did with the `storage` construct, we may want to enable versioning, encryption, etc.

I think to have such a feature it might better fit a new separate construct?

jlestel:
I’m agree, this is a better approach to use a new construct dedicated to this storage functionality.

I will work on the implementation to propose a PR in this direction.

mnapoli:
I'll close this one for now, but feel free to continue the discussion.

2 options to move forward:

- new "public-storage" construct
- new "public: true" option on existing "storage" construct

The second one might be the simplest, and it seems to make more sense. Note that in both cases the domain for this new construct would be different from another website construct.

Feel free to explain your use case in more details if that doesn't fit what you have in mind. So far what I understood is that you want to expose publicly assets that are generated by your application or uploaded by users.

Issue #188, opened 2022-04-21
I would like to be able to assign dead letter queue handler in lift


Setitch:
### Start from the Use-case

For now we can assign the handler for the queue, but when they fails messages goes to the dead letter queue - would be really nice if we could assign handler for that queue too

### Implementation Idea

deadHandler: <file>.<function> ?

mnapoli:
Could you detail what you would achieve with this setup?

Setitch:
Sorry for long delay.

I mean i would like to parse them by common lambda that would be handling dead deadletters

Setitch:
More details:
  i have a workflow, that moves the ,,ticket'' from one queue to another (they must be executed in order).
But when the ticket goes into deadletter queue, i want to grab it by lambda, try to reexecute it, and if its failing - remove from the deadletters but do some things - close the workflow for the ticket and so on.  Without providing the lambda on lift config for dead letter lambda-handler, its quite a task.

kamalnathlp:
@Setitch did you find any solution? @mnapoli any updates on the issue?

Setitch:
@kamalnathlp sadly not yet, which is real shame

fredericbarthelet:
DLQ url and arn are now exposed via variables. You can use those to attach a specific lambda handler.

Issue #128, opened 2021-11-02
Expose DLQ URL


pano-skylakis:
I'm currently retrieving the DLQ URL by referencing the logical ID of the DLQ that is generated after the first deployment.

For example in my serverless.yml:

```
environment:
  QUEUE_URL: ${construct:queue.queueUrl}
  DLQ_URL: !Ref queueDlqAB1234C5
```

Ideally, this value is exposed in on the queue when referencing the construct, so it could look like the following:

```
environment:
  QUEUE_URL: ${construct:queue.queueUrl}
  DLQ_URL: ${construct:queue.dlqUrl}
```

I've had a bit of a dig around and I think we can possibly achieve this by changing the following in [Queue.ts](https://github.com/getlift/lift/blob/d0a5ba7241a4b70da42812c0fc40fd34e8456bcf/src/constructs/aws/Queue.ts#L226)

```
variables(): Record<string, unknown> {
        return {
            queueUrl: this.queue.queueUrl,
            dlqUrl: dlq.queueUrl,
            queueArn: this.queue.queueArn,
        };
    }
```

t-richard:
Hi! Thanks for opening this issue.

In Lift, we like to start discussions from the use case (the why) instead of the solution (the how) ?

Could you describe the problem you're trying to solve with this ? 

PS: it was already asked in #63 but the use case was different from what is envisioned for the queue construct and led to a totally different discussion.

pano-skylakis:
Thanks for the response!

For sure, I'll elaborate.

In order to get access to the DLQ URL, I have to first deploy the queue and then either get the logical ID of the DLQ from the generated cloudformation template, or get the DLQ URL from the AWS console.

We end up with two different methods for accessing the queue URL and setting it as an environment variable, and getting the DLQ URL involves additional steps.

The reason for needing the DLQ URL is to manually push messages to the DLQ when an error occurs which we know will not resolve on a retry.

mnapoli:
> The reason for needing the DLQ URL is to manually push messages to the DLQ when an error occurs which we know will not resolve on a retry.

Ohh interesting 🤔  That makes sense, even though I never heard that use case before. I'll ask around, but that sounds reasonable to me? What do others think?

t-richard:
> The reason for needing the DLQ URL is to manually push messages to the DLQ when an error occurs which we know will not resolve on a retry.

I'm divided on this. _Maybe it makes sense in this specific domain use case you are facing but here's my two cents._

[...]

PR #275, opened 2022-11-03
Allow attachment of worker to Dead-letter queue

whambulance:
References Issue #188 

Changes aimed at allowing a worker to be assigned as a handler to the dead letter queue. Also publishing the dlq ARN to the yaml file - similar to how the main queue ARN 

Reasons:
* It was difficult to add a worker to a DLQ compared to how easy it was to add one to the main queue
* It would have required manually inserting the ARN into the YAML after having created the queue
* Allowing additional manipulation of the DLQ due to exposure of it's ARN in the YAML file

### DIFF
--- a/src/constructs/aws/Queue.ts
+++ b/src/constructs/aws/Queue.ts
@@ -37,13 +37,6 @@ const QUEUE_DEFINITION = {
             },
             additionalProperties: true,
         },
-        dlqWorker: {
-            type: "object",
-            properties: {
-                timeout: { type: "number" },
-            },
-            additionalProperties: true,
-        },
         maxRetries: { type: "number" },
         alarm: { type: "string" },
         batchSize: {
@@ -307,9 +300,9 @@ export class Queue extends AwsConstruct {
         const batchSize = this.configuration.batchSize ?? 1;
         const maximumBatchingWindow = this.getMaximumBatchingWindow();
 
-        // Override events for the queue worker
+        // Override events for the worker
         this.configuration.worker.events = [
-            // Subscribe the queue worker to the SQS queue
+            // Subscribe the worker to the SQS queue
             {
                 sqs: {
                     arn: this.queue.queueArn,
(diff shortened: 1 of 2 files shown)


mnapoli:
Hi, I'm curious why you would want to have a worker on the DLQ?

whambulance:
Hi @mnapoli - In my specific scenario, I'm using SQS to handle making api calls to a third party provider. If their API is unavailable for some reason, the messages will get pushed into the DLQ, and I need to be able to retry these failed messages at a second provider. This means changing up the content of the message.

[Here's another example from the AWS blog](https://aws.amazon.com/blogs/compute/using-amazon-sqs-dead-letter-queues-to-replay-messages/), written by one of their Senior solution architects.

Essentially, there exists use cases where simply pushing the failed messages back into the main queue isn't viable, and additional processing is required.

mnapoli:
Thanks for the explanation.

I think that falls outside of the scope of the construct, as it is right now. I don't see that feature being merged at the moment.

Maybe #128 is an alternative that has more chance of happening. Would that work for you?

whambulance:
As far as the base framework goes, exposing the URL allows you to send messages into the queue, but in order to consume messages from it I believe the ARN needs to be exposed?

Sadly I don't think this would be a viable solution - unless it's modified to expose the DLQ ARN too, similar to what I've done in this PR.

whambulance:
@mnapoli Updated to remove the dlqWorker construct, now only exposes dead letter queue URL & ARN

src/constructs/aws/Queue.ts
```diff
@@ -251,6 +252,10 @@ export class Queue extends AwsConstruct {
             description: `URL of the "${id}" SQS queue.`,
             value: this.queue.queueUrl,
         });
+        this.dlqArnOutput = new CfnOutput(this, "DlqArn", {
+            description: `ARN of the "${id}" SQS dead letter queue.`,
+            value: this.dlq.queueArn,
+        });
```
mnapoli, line 258: This CloudFormation output is not needed as far as I know.

src/constructs/aws/Queue.ts
```diff
@@ -119,6 +119,7 @@ export class Queue extends AwsConstruct {
     private readonly dlq: CdkQueue;
     private readonly queueArnOutput: CfnOutput;
     private readonly queueUrlOutput: CfnOutput;
+    private readonly dlqArnOutput: CfnOutput;
```
mnapoli, line 122: This variable becomes useless.

mnapoli (commented):
Thanks for the update, that looks good! I have added a few comments inline.

Could you also rebase on master? The tests are now passing on master, and this should allow us to test this PR.

src/constructs/aws/Queue.ts
mnapoli: Right, but the `${construct:jobs.dlqArn}` is transformed into a `!Ref` (cloudformation reference), it's not using the CF output.

The variables are exposed in the `variables()` method (not via CF outputs -> these are used for the CLI output).

whambulance: Ah - yes my bad, I believed it was the other way around. I've pushed a change that removes the initializer and the dlqArnOutput prop

fredericbarthelet (approved):

fredericbarthelet:
LGTM :) Merging and issuing release now ;) thanks @whambulance for the contribution !

Issue #277, opened 2022-11-17
Allow existing topic for DLQ Alarms


thejuan:
### Start from the Use-case

We have alerting infrastructure inplace that goes to more than just an email address, which we would like to re-use.

### Example Config

alarm: {topicArn} or ideally {topicName}

### Implementation Idea

If a topic is provided, use it rather than creating custom topics and email subs

fredericbarthelet:
Hey @thejuan , #265 is now fixed.
I included an exemple of what you're trying to achieve in the form of a unit test.
https://github.com/getlift/lift/blob/5ff580efe8f126eb8a6fa57082b8229f2467c0cb/test/unit/queues.test.ts#L624-L646
This should allow you to specify your own SNS topic as destination for the alarm.

I'm closing this issue in the meantime, please feel free to reopen if you have additional questions.

Thanks

Issue #265, opened 2022-09-14
Alarm extension not working properly on queues


codentacos:
### Description

I am attempting to create a CloudWatch alarm for a dead letter queue and it seems like this is not functioning as expected and may be a bug. Below is an example snippet of my queue construct. The idea here is to create a CloudWatch alarm that is only enabled in specific environments and the value is retrieved from `serverless.env.yml`

```
constructs:
  shipmentsUpdateQueue:
    type: queue
    fifo: true
    worker:
      handler: src/sqsHandlers/shipmentUpdateHandler.handle
    alarm: someemail@test.com
    extensions:
      alarm:
        Properties:
          ActionsEnabled:  ${self:custom.config.enableDeadLetterQueueAlarmActions}
```

Below is a snippet from the output CloudFormation template in the `.serverless` directory. The `ActionsEnabled` property is being set on the queue instead of the alarm as expected.
```
...
"shipmentsUpdateQueueDlqDF8B0233": {
      "Type": "AWS::SQS::Queue",
      "Properties": {
        "FifoQueue": true,
        "MessageRetentionPeriod": 1209600,
        "QueueName": "shipments-pr-8-shipmentsUpdateQueue-dlq.fifo",
        "ActionsEnabled": false,
...
```

I am also receiving the following error when I attempt to deploy to a PR stack
```
Error:
UPDATE_FAILED: shipmentsUpdateQueueDlqDF8B0233 (AWS::SQS::Queue)
Properties validation failed for resource shipmentsUpdateQueueDlqDF8B0233 with message:
#: extraneous key [ActionsEnabled] is not permitted
```

### How to Reproduce

This is my related serverless constructs configuration

```
constructs:
  shipmentsUpdateQueue:
    type: queue
    fifo: true
    worker:
      handler: src/sqsHandlers/shipmentUpdateHandler.handle
    alarm: someemail@test.com
    extensions:
      alarm:
        Properties:
          ActionsEnabled:  ${self:custom.config.enableDeadLetterQueueAlarmActions}
```

tpiere:
It looks like the `alarm` extension is pointing to the `dlq` which seems like a bug: https://github.com/getlift/lift/blob/163514966829cd5e1152011a3bdbe40e6d97b793/src/constructs/aws/Queue.ts#L283

PR #286, opened 2022-12-26
Allow existing SNS topic for DLQ SQS alarms

chrishanline:
References issue #277

We have existing alerting looking for `Alarm` and `OK` message posted to a specific SNS topic.

This change allows you to specify an existing SNS topic ARN for `alarm` on the DLQ SQS queue alarm construct. When an ARN that begins with `arn:aws:sns:` is present, that topic is used as an `AlarmAction` and `OKAction` instead of creating a new topic and email subscription.

No changes are made if an email address is specified for `alarm`.

Additionally adds checking for a bare-minimum of an ARN prefix or `@` in `alarm` and throwing an exception if neither is present.

### DIFF
--- a/src/constructs/aws/Queue.ts
+++ b/src/constructs/aws/Queue.ts
@@ -206,17 +206,31 @@ export class Queue extends AwsConstruct {
             ...encryption,
         });
 
-        const alarmEmail = configuration.alarm;
-        if (alarmEmail !== undefined) {
-            const alarmTopic = new Topic(this, "AlarmTopic", {
-                topicName: `${this.provider.stackName}-${id}-dlq-alarm-topic`,
-                displayName: `[Alert][${id}] There are failed jobs in the dead letter queue.`,
-            });
-            new Subscription(this, "AlarmTopicSubscription", {
-                topic: alarmTopic,
-                protocol: SubscriptionProtocol.EMAIL,
-                endpoint: alarmEmail,
-            });
+        if (configuration.alarm !== undefined) {
+            let alarmArn: string;
+            let alarmExistingTopic = undefined;
+            if (configuration.alarm.startsWith("arn:aws:sns:")) {
+                alarmArn = configuration.alarm;
+                alarmExistingTopic = true;
+            } else if (configuration.alarm.includes("@")) {
+                const alarmEmail = configuration.alarm;
+                const alarmTopic = new Topic(this, "AlarmTopic", {
+                    topicName: `${this.provider.stackName}-${id}-dlq-alarm-topic`,
+                    displayName: `[Alert][${id}] There are failed jobs in the dead letter queue.`,
+                });
+                new Subscription(this, "AlarmTopicSubscription", {
+                    topic: alarmTopic,
+                    protocol: SubscriptionProtocol.EMAIL,
+                    endpoint: alarmEmail,
+                });
+                alarmArn = alarmTopic.topicArn;
+                alarmExistingTopic = false;
+            } else {
+                throw new ServerlessError(
+                    `Invalid configuration in 'constructs.${this.id}': 'alarm' must be an SNS topic ARN that starts with 'arn:aws:sns:' or an email address with '@' present, '${configuration.alarm}' given.`,
+                    "LIFT_INVALID_CONSTRUCT_CONFIGURATION"
+                );
+            }
 
             const alarm = new Alarm(this, "Alarm", {
                 alarmName: `${this.provider.stackName}-${id}-dlq-alarm`,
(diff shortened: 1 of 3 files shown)


chrishanline:
@mnapoli or @fredericbarthelet any feedback or thoughts on merging this in?

src/constructs/aws/Queue.ts
```diff
@@ -206,17 +206,31 @@ export class Queue extends AwsConstruct {
...
-            });
-            new Subscription(this, "AlarmTopicSubscription", {
-                topic: alarmTopic,
-                protocol: SubscriptionProtocol.EMAIL,
-                endpoint: alarmEmail,
-            });
+        if (configuration.alarm !== undefined) {
+            let alarmArn: string;
+            let alarmExistingTopic = undefined;
+            if (configuration.alarm.startsWith("arn:aws:sns:")) {
```
fredericbarthelet, line 212: This would resolve only if user specify an hardcoded string as alarm topic. Any serverless resolvers used in place of an hardcoded string will fail this check.

fredericbarthelet:
Hi @chrishanline, thanks for proposing a PR for this feature !

We try to provide as simple configurations as possible in Lift, to avoid multiplication of configuration options, such as what you can find in the Serverless Framework current state.

With this in mind, and aware that everyone will have at some point specific requirements for its stack, we developed 2 features :

### Lift extensions

With extensions, you can manually modify generated Cloudformation to match your use case.
In this particular instance, the following code snippet will produce the expected behavior

```yml
constructs:
    my-queue:
        # ...
        alarm: dummy@dummy.com
        extensions:
            alarm:
                Properties:
                    AlarmActions: ["arn:aws:sns:us-east-1:123456789012:alarm-alerts"]
```

You can find out more about available extensions for the queue construct [here](https://github.com/getlift/lift/blob/master/docs/queue.md#extensions).

The downside is you'll still have a SNS topic and subscription being provisioned by Lift without actually being used.

### Eject

You can eject from Lift at anytime to regain control over the behavior of a single construct when your use case diverges too much from Lift original intent. You can find out more about ejecting in the corresponding documentation [here](https://github.com/getlift/lift#ejecting)

chrishanline:
hi @fredericbarthelet, thanks for the feedback.

when trying the extension, I was unsuccessful. I think I am running into this issue: https://github.com/getlift/lift/issues/265

when I do an eject and look at the cloudformation generated, the `AlarmActions` ends up on the `Queue` properties, not the alarm.

serverless.yml
```
constructs:
  print:
    type: queue
    fifo: true
    maxRetries: 2
    alarm: dummy@alarm.com
    extensions:
      alarm:
        Properties:
          AlarmActions: ["arn:aws:sns:us-east-1:123456789012:alarm-alerts"]
```

cloudformation from eject
```
  printDlq09F2B389:
    Type: 'AWS::SQS::Queue'
    Properties:
      FifoQueue: true
      MessageRetentionPeriod: 1209600
      QueueName: paperwork-printing-dev-print-dlq.fifo
      AlarmActions:
        - 'arn:aws:sns:us-east-1:123456789012:alarm-alerts'
    UpdateReplacePolicy: Delete
    DeletionPolicy: Delete
```

fredericbarthelet:
Indeed, would you be up to do a PR for #265 to solve this issue :) ?
It should be a matter of replacing the duplicate typo by a reference to the provisioned alarm

src/constructs/aws/Queue.ts
chrishanline: I just mocked up the following and it seemed to work fine per the cloudformation output from eject and did not throw a serverless error. But I may not understand what "serverless resolvers" means.

serverless.yml
```
custom:
  alarmSnsArn: arn:aws:sns:${aws:region}:${aws:accountId}:alarm-alerts

constructs:
  send:
    type: queue
    alarm: ${self:custom.alarmSnsArn}
```

cloudformation eject
```
  sendAlarmDFF3285F:
    Type: 'AWS::CloudWatch::Alarm'
    Properties:
      ComparisonOperator: GreaterThanThreshold
      EvaluationPeriods: 1
      AlarmActions:
        - 'arn:aws:sns:us-east-1:123456789012:alarm-alerts'
```

Is an example of a serverless resolver something like pulling in a CF output from another stack `alarm: !ImportValue another-stack-${self:provider.stage}-OutputArn`? this does throw a messy error currently it seems.

Is there anywhere else in lift this works I could use as an example to fix here?

mnapoli: It wouldn't work with, for example, `!Ref SomeTopic`

fredericbarthelet:
Hey @chrishanline, #265 is now fixed.
I included an exemple of what you're trying to achieve in the form of a unit test.
https://github.com/getlift/lift/blob/5ff580efe8f126eb8a6fa57082b8229f2467c0cb/test/unit/queues.test.ts#L624-L646
This should allow you to specify your own SNS topic as destination for the alarm.

I'm closing this PR in the meantime, please feel free to continue the conversation on #277 if you're still needing additional configuration options for this feature.

Thanks again for your time improving Lift :)

PR #287, opened 2022-12-30
Update comparison.md

CSFlorin:
SST has both [static website](https://docs.sst.dev/constructs/StaticSite) and [server-side rendering](https://docs.sst.dev/constructs/RemixSite)


mnapoli:
Thanks!

Issue #294, opened 2023-02-06
Add max concurrency support for AWS Lambda functions when using AWS SQS as an event source 


htxiong:
### Use-case description

12 Jan 2023 AWS introduced maximum concurrency of AWS Lambda functions when using Amazon SQS as an event source (see release [announcement here](https://aws.amazon.com/about-aws/whats-new/2023/01/aws-lambda-maximum-concurrency-amazon-sqs-event-source/) and [compute blog from AWS](https://aws.amazon.com/blogs/compute/introducing-maximum-concurrency-of-aws-lambda-functions-when-using-amazon-sqs-as-an-event-source/))

And Serverless started to support it around 3 weeks. [PR](https://github.com/serverless/serverless/issues/11676)

I suggest we should define max concurrency in Lift.

### Example Config

```
    alarm: engineering@xxx.com
    batchSize: 1
    maxRetries: 5
    maxConcurrency: 10
    worker:
```

### Implementation Idea

Add the configuration for `maximumConcurrency` and modify Queue.ts. ( e.g. after [this line](https://github.com/getlift/lift/blob/master/src/constructs/aws/Queue.ts#L305))

```
this.configuration.worker.events = [
            // Subscribe the worker to the SQS queue
            {
                sqs: {
                    arn: this.queue.queueArn,
                    batchSize: batchSize,
                    maximumBatchingWindow: maximumBatchingWindow,
                    maximumConcurrency: maximumConcurrency
                    functionResponseType: "ReportBatchItemFailures",
                },
            },
        ];
```

mnapoli:
That is 100% a good idea, thanks for opening this!

I'm wondering if `maximumConcurrency` shouldn't go one indentation lower, not in the `worker` but on the "queue" configuration itself, what do you think?

htxiong:
Yaaas, I will work on a PR.

htxiong:
@mnapoli can you review this PR for me?

fredericbarthelet:
Released in [1.25.0](https://github.com/getlift/lift/releases/tag/1.25.0)

PR #299, opened 2023-02-15
Add Max Concurrency configuration for AWS sqs.

htxiong:
this is the PR for issue https://github.com/getlift/lift/issues/294

### Use-case description

12 Jan 2023 AWS introduced maximum concurrency of AWS Lambda functions when using Amazon SQS as an event source (see release [announcement here](https://aws.amazon.com/about-aws/whats-new/2023/01/aws-lambda-maximum-concurrency-amazon-sqs-event-source/) and [compute blog from AWS](https://aws.amazon.com/blogs/compute/introducing-maximum-concurrency-of-aws-lambda-functions-when-using-amazon-sqs-as-an-event-source/))

And then Serverless started to support it in [late Jan 2023](https://github.com/serverless/serverless/pull/11678).

Add the configuration for maximumConcurrency  to the queue.

### Example Config

```
    batchSize: 1
    maxRetries: 5
    maxConcurrency: 10
    worker:
```


htxiong:
this is the PR for issue https://github.com/getlift/lift/issues/294

src/constructs/aws/Queue.ts
```diff
@@ -49,6 +49,11 @@ const QUEUE_DEFINITION = {
             minimum: 0,
             maximum: 300,
         },
+        maxConcurrency: {
+            type: "number",
+            minimum: 2,
+            maximum: 1000,
```
mnapoli, line 55: Is this really the maximum? The `1000` quota can be increased AFAIK, might be best not to limit it?

test/unit/queues.test.ts
```diff
@@ -89,6 +89,9 @@ describe("queues", () => {
                     "Fn::GetAtt": ["EmailsWorkerLambdaFunction", "Arn"],
                 },
                 MaximumBatchingWindowInSeconds: 0,
+                ScalingConfig: {
+                    MaximumConcurrency: 2,
+                },
```
mnapoli, line 94: This test should not change, because by default things should not change for users.

Instead, the `MaximumConcurrency` setting should be set only if they set it in `serverless.yml`.

I would suggest adding a new test that tests for `MaximumConcurrency`, and the current test here should not be modified. Does that make sense?

src/constructs/aws/Queue.ts
```diff
@@ -286,14 +291,22 @@ export class Queue extends AwsConstruct {
...
         return this.configuration.maxBatchingWindow ?? 0;
     }
 
+    private getMaximumConcurrency(): number {
+        return this.configuration.maxConcurrency ?? 2;
+    }
+
     private appendFunctions(): void {
-        // The default batch size is 1
-        const batchSize = this.configuration.batchSize ?? 1;
```
mnapoli, line 295: I don't think it's worth adding a method for this simple line.

mnapoli:
Could you update the documentation as well?

src/constructs/aws/Queue.ts
htxiong: ok, I will revert it, as i intended to make it consistent with other fields.

src/constructs/aws/Queue.ts
htxiong: there is a limit used by both AWS and Serverless, i do not think you could use any number larger than 1000.

test/unit/queues.test.ts
htxiong: its the default value for `MaximumConcurrency`, even you do not specify a default, it will use serverless default value thats 2.

htxiong:
@mnapoli addressed your comments. For the limit, AWS support only a number between 2 and 1000.

htxiong:
@mnapoli @j0k3r @kamalgill @jakejscott can someone review this PR again?

mnapoli:
@htxiong please don't ping users of the package unnecessarily.

htxiong:
> @htxiong please don't ping users of the package unnecessarily.
> 
> It seems the tests are failing though.

its weird, as it passed from my local. and i ran it again and it passed.

mnapoli:
Maybe it's necessary to update the CDK dependency?

htxiong:
i updated it to 2.65.0, it works as well. I do not know will it pass on CI if i push my change

htxiong:
Screenshot 2023-02-21 at 9 19 52 pm
it all passed.

babay123 (approved):

fredericbarthelet:
Hi @htxiong, I updated dev dependencies in #308
Could you rebase and repush your PR, including conflict solving in package.json
Hopefully, tests will be green and PR should be mergeable afterwards :)

Thanks

htxiong:
thanks @fredericbarthelet, its done. wait for someone to trigger the build and see will it pass all tests.

htxiong:
i merged the latest master and can someone trigger the CI job please?

mnapoli:
Done, tests are failing with serverless v2.36.

Is there a way to mark this test as skipped for v2 (older) serverless versions?

htxiong:
> Done, tests are failing with serverless v2.36.
> 
> Is there a way to mark this test as skipped for v2 (older) serverless versions?

thanks, there is a way to add condition for tests, i will check how can i create a test condition for serverless versions.

htxiong:
@mnapoli done, its not a nice way but it works as it.skipIf.

babay123:
> ### Example Config
> ```
>     alarm: engineering@xxx.com
>     batchSize: 1
>     maxRetries: 5
>     maxConcurrency: 10
>     worker:
> ```
Hello @htxiong 
What does alarm attribute used for ?

test/unit/queues.test.ts
```diff
@@ -226,6 +226,36 @@ describe("queues", () => {
         });
     });
 
+    it("allows changing the max concurrency", async () => {
+        const { cfTemplate } = await runServerless({
```
fredericbarthelet, line 230: You can retrieve serverless current version much more easily :
```suggestion
        const { cfTemplate, serverless } = await runServerless({
```

and then, you have access to `serverless.version` that outputs the version of the framework :)

fredericbarthelet (changes requested):
@htxiong you have an easier method to retrieve the serverless version. We should be good to merge after that. Thanks for your persistence :) !

htxiong:
@fredericbarthelet thanks and i changed it. please review it again and trigger the CI jobs.

fredericbarthelet (approved):

fredericbarthelet:
Thanks for the contribution @htxiong :)

PR #302, opened 2023-02-16
Correct misleading deduplication documentation

seantcanavan:
Content based deduplication MUST be explicitly disabled to support sending deduplication IDs. I've been hunting down a bug for months trying to figure out what's going wrong. Setting a deduplication ID with content based deduplication enabled does nothing - it still uses the body.


mnapoli:
Thank you!

PR #306, opened 2023-02-23
Simplify server-side-website with the new AllViewerExceptHostHeader policy

### LINKED ISSUE #144: Cache Policy for Server-Side-Website

### LINKED ISSUE #229: Allow Use of AllViewer Managed Origin Request Policy
### Start from the Use-case

It is easy to bump up against the limit of ten white-listed headers to pass through CloudFront for `server-side-website` apps. However, AWS has a managed policy, [AllViewer](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-origin-request-policies.html#managed-origin-request-policies-list), that passes through all headers, cookies, and query strings. I recently had an issue getting a Laravel site that uses Livewire to function properly, as neither project makes clear in their docs what headers they rely on for their security features.

### Example Config

```
constructs:
    website:
        # ...
        forwardedHeaders:
            - Accept
            - Accept-Language
            # ...
            - X-Custom-Header10
            - X-Custom-Header11 # Oops, too many! Assuming I even know what headers to whitelist.
```

### Implementation Idea

I suggest that the `server-side-website` configuration default to use of the `AllViewer` (id:216adef6-5c7f-47e4-b989-5492eafa07d3) managed origin request policy in the absence of the `forwardedHeaders` directive. If the `forwardedHeaders` directive is present, that will override the new default behavior and will create a new origin request policy the same as current behavior.

This change requires no changes to the configuration file, but it does have implicit behavior, which is not ideal.

As an alternative, we could add the following directive: `passAllHeadersToOrigin` with a default of `true`. The value of `true` will signal the use of the `AllViewer` managed origin request policy. A value of `false` will indicate the need for the `forwardedHeaders` directive.

I looked over the other four managed origin request policies, and none of them seemed useful or applicable to the `server-side-website` configuration, so as far as I can tell, the only meaningful options are managed (`AllViewer`) or custom.

I'm not familiar with Typescript, but on cursory examination, it shouldn't be a huge change to the code that actually generates the CloudFormation script. [Line 120 of ServerSideWebsite.ts](https://github.com/getlift/lift/blob/94387e6e1d004fe9360a47f947333f2ee80b3af6/src/constructs/aws/ServerSideWebsite.ts#L120) is where we'd put a check to determine managed or custom origin request policy.

It appears that we could instead say:
```ts
if (configuration.passAllHeadersToOrigin) {
    const backendOriginPolicy = OriginRequestPolicy.ALL_VIEWER;
}
else {
    // Current custom forwarded headers code.
}
```
based on my understanding of line [66 in origin-request-policy.ts](https://github.com/aws/aws-cdk/blob/74318c7d22bfc00de9e005f68a0a6aaa58c7db39/packages/%40aws-cdk/aws-cloudfront/lib/origin-request-policy.ts#L66).

Unless there are some security issues I'm not aware of, I think this would be the best way to turn CloudFront into a dumb pipe for the back-end server.

### LINKED ISSUE #291: AWS limit per account
### Start from the Use-case

When you deploy a server-side website with lift it will create a new Cache policy, a new Origin request policy and a CloudFront function every time; and most of times these resources are the same.

AWS has a limit of 20 Cache policies and Origin request policies and 100 CloudFront Functions per AWS account; and these limits are not increaseble ([more info here](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/cloudfront-limits.html)).

When you have a really huge amount of applications running in an AWS account, this can be a big limitation.

To solve this problem, one could envisage reusing these resources by specifying their IDs in the configuration of the website construct.

### Example Config

```yaml
constructs:
    website:
        # ...
        backendOriginPolicy: "..."
        backendCachePolicy: "...."
        cloudfrontFunctionArn: "arn:aws:cloudfront::..."
```

### Implementation Idea

[Pull request here](https://github.com/getlift/lift/pull/290)

mnapoli:
CloudFront added a new `AllViewerExceptHostHeader` policy: https://twitter.com/cristiangraz/status/1628585607479050240

This is exactly what we needed!

We can get rid of:

- our custom "origin policy" (it was needed to avoid forwarding the Host header)
- our custom "cache policy" (it was needed to forward the Authorization header)

We gain:

- less code for us
- less config for users 🎉
- less resources to deploy, meaning faster deployments 🎉
- users don't have to choose or configure the headers they want to forward 🎉
- users are no longer limited by the limit of 10 forwarded headers 🎉

If you were setting the `forwardedHeader` config in `serverless.yml` to add more headers to forward, you can now safely remove this option:

```yaml
constructs:
    website:
        # ...

        # REMOVE THIS 👇
        forwardedHeaders:
            - Accept
            - Accept-Language
            # ...
            - X-Custom-Header
```

We also finally solve 100% properly #144

Fixes #144
Fixes #229
Fixes #291
and so many more in bref.sh 🎉

### DIFF
--- a/docs/server-side-website.md
+++ b/docs/server-side-website.md
@@ -335,34 +335,13 @@ Applications are of course free to catch errors and display custom error pages. 
 
 ### Forwarded headers
 
-By default, CloudFront is configured to forward the following HTTP headers to the backend running on Lambda:
+> **Note**:
+> 
+> In previous Lift versions, the headers forwarded to your Lambda backend were limited. You were able to add new headers via the `forwardedHeaders` option.
+> 
+> [This is no longer the case](https://twitter.com/cristiangraz/status/1628585607479050240): **all headers are now always forwarded**.
 
-- `Accept`
-- `Accept-Language`
-- `Authorization`
-- `Content-Type`
-- `Origin`
-- `Referer`
-- `User-Agent`
-- `X-Forwarded-Host`
-- `X-Requested-With`
-
-Why only this list? Because CloudFront + API Gateway requires us to define explicitly the list of headers to forward. It isn't possible to forward _all_ headers.
-
-To access more headers from the client (or [from CloudFront](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-cloudfront-headers.html)), you can redefine the list of headers forwarded by CloudFront in `forwardedHeaders`:
-
-```yaml
-constructs:
-    website:
-        # ...
-        forwardedHeaders:
-            - Accept
-            - Accept-Language
-            # ...
-            - X-Custom-Header
-```
-
-CloudFront accepts maximum 10 headers.
+All headers are forwarded to your Lambda backend. If you used the `forwardedHeaders` option in the past, you can safely remove it.
 
 ## Extensions
 
--- a/src/constructs/aws/ServerSideWebsite.ts
+++ b/src/constructs/aws/ServerSideWebsite.ts
@@ -1,20 +1,13 @@
 import type { CfnBucket } from "aws-cdk-lib/aws-s3";
 import { Bucket } from "aws-cdk-lib/aws-s3";
-import type { CfnDistribution } from "aws-cdk-lib/aws-cloudfront";
+import type { CfnDistribution, IOriginRequestPolicy } from "aws-cdk-lib/aws-cloudfront";
 import {
     AllowedMethods,
-    CacheCookieBehavior,
-    CacheHeaderBehavior,
     CachePolicy,
-    CacheQueryStringBehavior,
     Distribution,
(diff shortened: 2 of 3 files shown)


kevincerro:
Great news 🚀

buddhaCode:
Does this also solve the issue, that u have to name/list all custom headers like `X-Livewire` and custom headers for CSRF token in Laravel?

mnapoli:
@buddhaCode yeeeessss!

esimonetti:
:heart: thank you!!!!

PR #312, opened 2023-03-01
Rewrite disable auto permission test

fredericbarthelet:
Serverless v3.28.0 updates permissions given to lambda to be able to use a new logs:TagResource action on Cloudwatch groups. Rewriting the test to ensure no permission regarding s3 are added to lambda instead of checking exact Action match assertions, in order to have tests passing both for serverless v2 and latest v3

PR made on Serverless v3.28.0 that introduce this change : https://github.com/serverless/serverless/pull/11766


mnapoli:
Awesome!

PR #321, opened 2023-04-20
Gracefully handle redirectToMainDomain with only a single domain

aran112000:
We've run into issues with dynamic configs which are stage-dependent - different stages have different domain(s).

This change allows the directive to be present but will silence its behaviour where only a single domain has been specified.

### DIFF
--- a/src/classes/cloudfrontFunctions.ts
+++ b/src/classes/cloudfrontFunctions.ts
@@ -2,10 +2,7 @@ import ServerlessError from "../utils/error";
 
 export function redirectToMainDomain(domains: string[] | undefined): string {
     if (domains === undefined || domains.length < 2) {
-        throw new ServerlessError(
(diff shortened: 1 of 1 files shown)


mnapoli:
The tests are failing, any idea why?

aran112000:
> The tests are failing, any idea why?

I believe these should all be good to go now :)

src/constructs/aws/ServerSideWebsite.ts
```diff
@@ -372,7 +372,7 @@ export class ServerSideWebsite extends AwsConstruct {
     private createRequestFunction(): cloudfront.Function {
         let additionalCode = "";
 
-        if (this.configuration.redirectToMainDomain === true) {
+        if (this.configuration.redirectToMainDomain === true && this?.domains && this.domains.length > 1) {
```
mnapoli, line 375: The build still fails, could it be that `this?` is invalid because `this` cannot be null?

mnapoli:
Thanks!

Issue #331, opened 2023-05-03
SinglePageApp should support .wasm files


AtteR:
### Start from the Use-case

Hello,

Currently CloudFront request function redirects .wasm files to /index.html (https://github.com/getlift/lift/blob/master/src/constructs/aws/SinglePageApp.ts#L48)

This makes the .wasm files load as HTML, which means that .wasm file is not loaded properly.

Adding wasm to REDIRECT_INDEX seems to fix this problem.

### Implementation Idea

Add wasm to REDIRECT_REGEX:

var REDIRECT_REGEX = /^[^.]+$|\.(?!(css|gif|ico|jpg|jpeg|js|png|txt|svg|woff|woff2|ttf|map|json|webp|xml|pdf|webmanifest|avif|wasm)$)([^.]+$)/;

This seems to fix the issue.

PR #332, opened 2023-05-04
Add wasm to redirect regex getlift/lift#331

AtteR:
https://github.com/getlift/lift/issues/331

Currently wasm files are not loaded properly as the CloudFront function redirects them to .html. Adding wasm to redirect regex fixes this and wasm files are loaded properly.


mnapoli:
Hi @AtteR, tests are failing, I think you will need to update them as well?

AtteR:
Hi @mnapoli ah, yes, should be fixed now :)

mnapoli:
Thank you!

PR #336, opened 2023-06-02
Allow the usage under Node 20.

dnp1:
Allow the usage under Node 20.

The CI worked here:
https://github.com/dnp1/lift/actions/runs/5155713648

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,13 +17,14 @@ jobs:
                     - 14
                     - 16
                     - 18
+                    - 20
                 serverless:
                     - "2.36.0"
                     - "2"
                     - "latest"
         steps:
             - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -17,13 +17,14 @@ jobs:
...
                     - "2.36.0"
                     - "2"
                     - "latest"
         steps:
             - uses: actions/checkout@v2
-            - uses: actions/setup-node@v2
+            - uses: actions/setup-node@v3
               with:
                   node-version: ${{ matrix.node }}
             - uses: actions/cache@v2
```
dnp1, line 30: @mnapoli maybe here we could use the built-in cache of `actions/setup-node@v3`

mnapoli: Indeed that would be great as a separate PR!

PR #337, opened 2023-06-06
SQS DLQ alarm should treat missing data as not breaching

chrishanline:
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-monitoring-using-cloudwatch.html

> When an Amazon SQS queue is inactive for more than six hours, the Amazon SQS service is considered asleep and stops delivering metrics to the CloudWatch service. Missing data, or data representing zero, can't be visualized in the CloudWatch metrics for Amazon SQS for the time period that your Amazon SQS queue was inactive.

Because of this, the DLQ alarms currently created by lift will show `Insufficient data` most of the time when the queue they monitor are empty _and inactive_.

dlq cw alarms

This PR sets `treatMissingData` to `notBreaching` on the alarm vs the current CDK default of `missing`. `notBreaching` will allow an inactive (and thus empty) SQS queue to show `Ok` when browsing alarms.


mnapoli:
Thank you!

PR #340, opened 2023-06-11
fix: adds `dynamodb:ConditionCheckItem` permission to dynamodb construct

### LINKED ISSUE #339: DynamoDB IAM policy doesn't include `dynamodb:ConditionCheck`
I noticed whilst deploying recently that a feature I was using for DynamoDB using `ConditionCheck` for write transactions is not allowed out of the box with lift. I can't see any documentation as to why this is the case and believe it is an oversight as it is a useful feature when writing transacted statements with dynamodb, see this aws article for details:

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transaction-apis-iam.html

The offending line is: https://github.com/getlift/lift/blob/71b15804ec24bceda35e8a1cf6fe48f2c32567eb/src/constructs/aws/DatabaseDynamoDBSingleTable.ts#LL88C29-L88C29

I believe simply adding `"dynamodb:ConditionCheckItem"` would enable this feature without having to modify the default permissions manually.

cmcnicholas:
Closes #339 

Without this IAM permission, transacted writes in dynamodb using the `ConditionCheck` feature are disallowed.

See:

https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_ConditionCheck.html

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transaction-apis-iam.html

Example usage:

After this change you would be able to make use of ConditionCheck in transacted writes e.g.

```ts
client.send(
  new TransactWriteCommand({
    TransactItems: [
      // checks an item condition is successful
      {
        ConditionCheck: {
          TableName: environment.tableName,
          Key: {
            PK: 'somepk',
            SK: 'somesk',
          },
          ConditionExpression: 'attribute_exists(PK)',
        },
      },
      // and writes a record in 1 transaction ensuring consistency
      {
        Put: {
          TableName: environment.tableName,
          Item: someItem,
        },
      }
    ]
  }));
```

### DIFF
--- a/src/constructs/aws/DatabaseDynamoDBSingleTable.ts
+++ b/src/constructs/aws/DatabaseDynamoDBSingleTable.ts
@@ -86,6 +86,7 @@ export class DatabaseDynamoDBSingleTable extends AwsConstruct {
                     "dynamodb:DeleteItem",
                     "dynamodb:BatchWriteItem",
                     "dynamodb:UpdateItem",
+                    "dynamodb:ConditionCheckItem",
                 ],
                 [this.table.tableArn, Stack.of(this).resolve(Fn.join("/", [this.table.tableArn, "index", "*"]))]
             ),


mnapoli:
Thanks!

PR #341, opened 2023-06-14
Add in method setting for webhooks

SludgeGirl:
Some webhooks utilise PUT instead of strictly POST, in order to support them I've added in the ability to specify which method the webhook should allow in order to allow for these cases

Let me know if you want any changes to this, it's my first time touching the project so suggestions are more than welcome

### DIFF
--- a/src/constructs/aws/Webhook.ts
+++ b/src/constructs/aws/Webhook.ts
@@ -28,7 +28,7 @@ const WEBHOOK_DEFINITION = {
         insecure: { type: "boolean" },
         path: { type: "string" },
         eventType: { type: "string" },
-        method: { type: "string" },
+        method: { enum: ["POST", "PUT", "PATCH"] },
     },
     required: ["path"],
     additionalProperties: false,
--- a/src/constructs/aws/Webhook.ts
+++ b/src/constructs/aws/Webhook.ts
@@ -28,12 +28,14 @@ const WEBHOOK_DEFINITION = {
         insecure: { type: "boolean" },
         path: { type: "string" },
         eventType: { type: "string" },
+        method: { type: "string" },
     },
     required: ["path"],
     additionalProperties: false,
 } as const;
 const WEBHOOK_DEFAULTS = {
     insecure: false,
+    method: "POST",
 };
 
 type Configuration = FromSchema<typeof WEBHOOK_DEFINITION>;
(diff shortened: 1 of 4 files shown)


SludgeGirl:
Sorry for the extra force, after a bit more time to think on it I wasn't quite happy with my handling of getHttpMethod, I'm still quite, getEndpointPath only returns the first instance of the output it finds in the logs, this leads to only finding the first method set. 

Due to the promise required outputs() from AwsConstruct I can't just return this.methods.join(', ') nor can I use the stack output due to it only returning the first instance

I'd very happily get your opinion on hows best to handle this

SludgeGirl:
Yep that'd make a lot of sense, I've just got those changes pushed up

mnapoli:
@fredericbarthelet WDYT?

mnapoli:
I'd like to get @fredericbarthelet's opinion on this. I think this PR makes sense.

In any case we'd need tests to cover the new feature.

src/constructs/aws/Webhook.ts
```diff
@@ -28,12 +28,14 @@ const WEBHOOK_DEFINITION = {
         insecure: { type: "boolean" },
         path: { type: "string" },
         eventType: { type: "string" },
+        method: { type: "string" },
```
fredericbarthelet, line 31: I'd be a bit more precise here, in order to avoid HTTP verbs without body - API Gateway would not throw an error for a `GET` method, but integration with Eventbridge would crash as no body payload would be accessible : 
```suggestion
        method: { enum: ["POST", "PUT", "PATCH"] },
```

fredericbarthelet (changes requested):
LGTM :) Thanks @SludgeGirl !
Could you update the documentation at https://github.com/getlift/lift/blob/1e786ed4535de2d004b71b458b55cf9f2e39a9af/docs/webhook.md?plain=1#L69 and include corresponding tests in test/unit/webhooks.test.ts ?

Once added, we'll be merging this feature !

src/constructs/aws/Webhook.ts
SludgeGirl: My latest changes have switched over to that!

SludgeGirl:
I've rebased everything, added in the changes to the docs and unit tests. Let me know if you want anything further done!

mnapoli (approved):
I'll let @fredericbarthelet review and approve

fredericbarthelet (approved):

fredericbarthelet:
Thanks for your contribution @SludgeGirl :) ! Will be releasing a new Lift version right away for you to be able tu use this feature.

PR #342, opened 2023-06-20
When we have custom domain(s), always redirect to them

aran112000:
This ensures API Gateway and Cloudfront URLs all redirect to the primary domain - this helps prevent duplicating sites and getting them indexed across multiple hosts accidentally

### DIFF
--- a/src/classes/cloudfrontFunctions.ts
+++ b/src/classes/cloudfrontFunctions.ts
@@ -1,7 +1,8 @@
 import ServerlessError from "../utils/error";
 
 export function redirectToMainDomain(domains: string[] | undefined): string {
-    if (domains === undefined || domains.length < 2) {
+    if (domains === undefined) {
+        // No custom domains specified
         return "";
     }
 


mnapoli:
Hi Aran, not sure I get this: the API Gateway URL will not invoke the CloudFront function, right?

aran112000:
Sorry, to explain more clearly, when we deploy a website with a custom domain, we end up with URLs for:
 - Cloudfront: https://xxxxx.cloudfront.net	
 - API Gateway: https://xxxxxx.execute-api.eu-west-2.amazonaws.com
 - + any custom domains we've mapped

This PR would allow this option to redirect ALL hosts to the configured primary domain.

Without this, if a search engine managed to crawl these URLs, all of these versions would show the same content unless handled within the application logic (easy to forget and end up with nasty SEO implications as a result).

mnapoli:
Ahh ok, I guess this would work for the cloudfront.net URL. But I meant it won't work for the API Gateway URL, right?

aran112000:
I was looking to solve Cloudfront initially and just thought (wrongly) this would just work with APIGW too, but of course not now I think through it properly.

Issue #266, opened 2022-09-15
Add SNS subscriptions to queues


jesusch:
I assum this could be a common usecase

```
constructs:
  my-queue:
    type: queue
    subscriptions:
    - arn:aws:sns:us-east-1:1234567890:my-topic
    - arn:aws:sns:eu-west-1:1234567890:my-othertopic
```

mnapoli:
What would be the use case?

jesusch:
Subscribe sqs queues to sns topics

ryanwalters:
> What would be the use case?

A common pattern, usually called "fanout", is to subscribe SQS queues to SNS topics. AWS documentation has a basic example of the pattern here: https://aws.amazon.com/getting-started/hands-on/send-fanout-event-notifications/

Here's an example of what the CloudFormation configuration looks like (there's probably a CDK example out there too): https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-sns-subscription.html#aws-resource-sns-subscription--examples--Create_a_subscription_with_optional_attributes--yaml

Note the protocol set to `sqs`.

cmcnicholas:
> > What would be the use case?
> 
> A common pattern, usually called "fanout", is to subscribe SQS queues to SNS topics. AWS documentation has a basic example of the pattern here: https://aws.amazon.com/getting-started/hands-on/send-fanout-event-notifications/
> 
> Here's an example of what the CloudFormation configuration looks like (there's probably a CDK example out there too): https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-sns-subscription.html#aws-resource-sns-subscription--examples--Create_a_subscription_with_optional_attributes--yaml
> 
> Note the protocol set to `sqs`.

Exactly this, I'm experiencing this on a project at the moment and have seen this used before in a few places, it's quite common for allowing resilient simplified async processing of independent pipelines e.g. 

![image](https://github.com/getlift/lift/assets/41394833/5e05ff07-330b-49b2-a04d-54a9b507fa5c)

cmcnicholas:
I've had a stab at implementing this, see #344 

I would appreciate any help simplifying or helping to resolve the issue about having to use ARN strings instead of Refs.

cmcnicholas:
sorry to chase up on this, after doing a little bit of digging I still can't find the "plugin way" to resolve refs in the linked PR e.g.

```yaml
constructs:
  ## adds a simple queue and unfiltered subscription
[...]

PR #344, opened 2023-07-02
feat: adds subscription support

cmcnicholas:
This PR adds support for subscriptions in Queue constructs, see: #266 

There are quite a few use cases for this but a common pattern is "fanout" https://aws.amazon.com/getting-started/hands-on/send-fanout-event-notifications/ used to distribute messages to multiple queues that have different processing concerns/domains.

I need some help on this PR as it isn't "oven ready"...

Specifically Line 313 `Queue.ts` uses `Topic.fromTopicArn` to get a reference to the topic that should be subscribed to.

The `subscription.topic` is a string so at this point MUST be a resolved ARN. The following `serverless.yml` works fine:

```yml
service: lift-test
configValidationMode: error

plugins:
  - serverless-lift

provider:
  name: aws
  region: eu-west-1
  versionFunctions: false

resources:
  Resources:
    MyTopic:
      Type: 'AWS::SNS::Topic'
      Properties:
        TopicName: SuperTopic

constructs:
  ## adds a simple queue and unfiltered subscription
  my-queue:
    type: queue
    worker:
      handler: worker.handler
    subscriptions:
      - topic: arn:aws:sns:eu-west-1:XXXXXXXXX:SuperTopic

  ## more complex: adds multiple filtered subscriptions
  my-queue:
    type: queue
    worker:
      handler: worker.handler
    subscriptions:
      - topic: arn:aws:sns:eu-west-1:XXXXXXXXX:SuperTopic
      - topic: arn:aws:sns:eu-west-1:XXXXXXXXX:SuperTopic
        filters:
          - attribute: yyy
            allows:
              - hello
              - world
      - topic: arn:aws:sns:eu-west-1:XXXXXXXXX:SuperTopic
        filters:
          - attribute: xxx
            denied:
              - other
              - world
```

What I would want this to do though is to be able to accept ref/variables so you could avoid having to use an ARN as string.

```yml
service: lift-test
configValidationMode: error

plugins:
  - serverless-lift

provider:
  name: aws
  region: eu-west-1
  versionFunctions: false

resources:
  Resources:
    MyTopic:
      Type: 'AWS::SNS::Topic'
      Properties:
        TopicName: SuperTopic

constructs:
  ## adds a simple queue and unfiltered subscription
  my-queue:
    type: queue
    worker:
      handler: worker.handler
    subscriptions:
      - topic: !Ref MyTopic # this doesn't work
```

I understand why this doesn't work, lift config is expecting a string but I really would like to "resolve" any dynamic references for this to be automatic. My knowledge of CDK/Lift is limited in this area so would appreciate anyone pointing me in the right direction 👋

### DIFF
--- a/src/constructs/aws/Queue.ts
+++ b/src/constructs/aws/Queue.ts
@@ -4,7 +4,7 @@ import { Queue as CdkQueue, QueueEncryption } from "aws-cdk-lib/aws-sqs";
 import type { FromSchema } from "json-schema-to-ts";
 import type { CfnAlarm } from "aws-cdk-lib/aws-cloudwatch";
 import { Alarm, ComparisonOperator, Metric, TreatMissingData } from "aws-cdk-lib/aws-cloudwatch";
-import { Subscription, SubscriptionProtocol, Topic } from "aws-cdk-lib/aws-sns";
+import { Subscription, SubscriptionFilter, SubscriptionProtocol, Topic } from "aws-cdk-lib/aws-sns";
 import type { AlarmActionConfig } from "aws-cdk-lib/aws-cloudwatch/lib/alarm-action";
 import type { Construct as CdkConstruct } from "constructs";
 import type { CfnResource } from "aws-cdk-lib";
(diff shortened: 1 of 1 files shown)


cmcnicholas:
OK further to the above I have managed to get the following to work:

```ts
const fn = Fn.getAtt("MyTopic", "TopicArn");
const topic = Topic.fromTopicArn(this, "SomeId", fn.toString());
topic.addSubscription(new SqsSubscription(this.queue, { filterPolicy, rawMessageDelivery: true }));
```

if there a friendly way to serialise/deserialise Fn.getAtt("MyTopic", "TopicArn") into the lift config?

Examples show use of the following:

```yaml
constructs:
  one-of-them-queues:
    type: queue
    worker:
      handler: worker.handler
    subscriptions:
      - topicRef:
          Fn::GetAtt:
            - MyTopic
            - TopicArn
```

But when doing so I get `An ARN must have at least 6 elements, not 4`

cmcnicholas:
This PR is now ready :)

PR #349, opened 2023-07-26
Fix #348 topic display name too long

### LINKED ISSUE #348: Display Name Length Restriction in Queues
### Description

The current Queue generation allows Alarm Topics to be created with the `displayName` property exceeding 100 characters (a restriction by the AWS API). This results in failed cloudformation deployment with the error message:

```
Resource handler returned message: "Invalid parameter: Attributes Reason: DisplayName (Service: Sns, Status Code: 400, Request ID: 22d6afac-465f-5167-8e0d-f297e231f7b2)" (RequestToken: 660cee2b-d73b-49d0-879c-eb88de339e79, HandlerErrorCode: GeneralServiceException)
```

### How to Reproduce

Create a queue with alarm that has a construct name ~60+ characters

### Additional Information

_No response_

cmcnicholas:
fixes #348 where the `displayName` property exceeds the aqs model max length. See associated bug report for error message.

this PR tries to "pretty print" the display name in situations where the id is long enough to make the displayName > 100 chars

### DIFF
--- a/src/constructs/aws/Queue.ts
+++ b/src/constructs/aws/Queue.ts
@@ -232,9 +232,19 @@ export class Queue extends AwsConstruct {
 
         const alarmEmail = configuration.alarm;
         if (alarmEmail !== undefined) {
+            // generate the display name, AWS restriction is 100 chars
+            let displayName = `[Alert][${id}] failed jobs in dlq.`;
+            if (displayName.length > 100) {
+                // if the length is too long then try to pretty print
+                displayName = `[Alert][${id.substring(
+                    0,
+                    id.length - (displayName.length - 100 - 3)
+                )}...] failed jobs in dlq.`;
+            }
+
             const alarmTopic = new Topic(this, "AlarmTopic", {
                 topicName: `${this.provider.stackName}-${id}-dlq-alarm-topic`,
-                displayName: `[Alert][${id}] There are failed jobs in the dead letter queue.`,
+                displayName,
             });
             new Subscription(this, "AlarmTopicSubscription", {
                 topic: alarmTopic,


src/constructs/aws/Queue.ts
```diff
@@ -232,9 +232,19 @@ export class Queue extends AwsConstruct {
...
         if (alarmEmail !== undefined) {
+            // generate the display name, AWS restriction is 100 chars
+            let displayName = `[Alert][${id}] failed jobs in dlq.`;
+            if (displayName.length > 100) {
+                // if the length is too long then try to pretty print
+                displayName = `[Alert][${id.substring(
+                    0,
+                    id.length - (displayName.length - 100 - 3)
+                )}...] failed jobs in dlq.`;
+            }
```
mnapoli, line 243: Thanks for the PR!

I think we can get rid of the `if` and always truncate: if the max length is not reached then we'll have the full name.

src/constructs/aws/Queue.ts
```diff
@@ -232,9 +232,19 @@ export class Queue extends AwsConstruct {
...
         const alarmEmail = configuration.alarm;
         if (alarmEmail !== undefined) {
+            // generate the display name, AWS restriction is 100 chars
+            let displayName = `[Alert][${id}] failed jobs in dlq.`;
+            if (displayName.length > 100) {
+                // if the length is too long then try to pretty print
+                displayName = `[Alert][${id.substring(
+                    0,
+                    id.length - (displayName.length - 100 - 3)
+                )}...] failed jobs in dlq.`;
```
mnapoli, line 242:
```suggestion
                )}...] Failed jobs in the DLQ.`;
```

src/constructs/aws/Queue.ts
mnapoli: The first option is what I meant 👍

PR #353, opened 2023-08-16
Add distribution id to server side website

evo-chris:
Output the Cloudfront Distribution Id so can be used for IAM policies

### DIFF
--- a/docs/server-side-website.md
+++ b/docs/server-side-website.md
@@ -177,6 +177,26 @@ resources:
                 # ...
 ```
 
+- `distributionId`: the Cloudfront Distribution Id
+
+This can be used to configure a Cloudfront Invalidation Policy
+
+```yaml
+constructs:
+    website:
+        type: server-side-website
+        # ...
+iam:
+  role:
+    statements:
+        - Effect: Allow
+          Action:
+            - 'cloudfront:CreateInvalidation'
+          Resource:
+            - 'arn:aws:cloudfront::${aws:accountId}:distribution/${construct:website.distributionId}'
+          # ...
+```
+
 ## Commands
 
 `serverless deploy` deploys everything configured in `serverless.yml` and uploads assets.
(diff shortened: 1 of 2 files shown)


mnapoli:
Hi, could you add some tests to cover the new feature?

robchett:
Hi @mnapoli, I can't see anywhere where there are tests on the existing outputs, `url`, `cname` & `assetsBucketName` or see a way to verify they are accessible in the outputs of `runServerless`

I've tried to see if I could infer they exist but referencing them in a template

```
        const { cfTemplate, computeLogicalId } = await runServerless({
            command: "package",
            config: Object.assign(baseConfig, {
                provider: {
                    name: "aws",
                    iam: {
                        role: {
                            statements: [{
                                Effect: "Allow",
                                Action: [
                                    'cloudfront:CreateInvalidation'
                                ],
                                Resource: [
                                    'arn:aws:cloudfront::${aws:accountId}:distribution/${construct:backend.distributionId}'
                                ]
                            }]
                        }
                    }
                },
                constructs: {
                    backend: {
                        type: "server-side-website",
                        assets: {
                            "/assets/*": "public",
                        },
                    },
                },
            }),
        });
```
but while it will error if I mess with the object itself it doesn't complain about an invalid reference. 

I've also tried using a fixture

```
service: website
configValidationMode: error

provider:
    name: aws
    iam:
        role:
            statements:
                - 
                    Effect: Allow                
                    Action:
                        - 'cloudfront:CreateInvalidation'
                    Resource:
                        - 'arn:aws:cloudfront::${aws:accountId}:distribution/${construct:website.distributionId}'
                        
functions:
    http:
        handler: handler.handler

constructs:
    backend:
        type: server-side-website
        assets:
            "/assets/*": public
        errorPage: error.html
```
but the variable are not replaced 

```
 {
                      Effect: 'Allow',
                      Action: [ 'logs:PutLogEvents' ],
                      Resource: [
                        {
                          'Fn::Sub': 'arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/test-serverSideWebsite-2ol0dc8r-dev*:*:*'
                        }
                      ]
                    },
                    {
                      Effect: 'Allow',
                      Action: [ 'cloudfront:CreateInvalidation' ],
                      Resource: 'arn:aws:cloudfront::${aws:accountId}:distribution/${construct:website.distributionId}'
                    }
```

Any thoughts on how to go about testing this?

PR #360, opened 2023-09-19
Feat/add origin spa

cdominguezg:
I need to add the origins in order to use API Gateway resources. Tested with extensions but didn't work, so I've implemented another fields to do that.

I've added a new test and also tested in an internal project as a PoC and it works as expected. I updated the docs.

### DIFF
--- a/docs/single-page-app.md
+++ b/docs/single-page-app.md
@@ -17,6 +17,22 @@ constructs:
     landing:
         type: single-page-app
         path: public
+        origins:
+            - path: staging/
+              pathPattern: api/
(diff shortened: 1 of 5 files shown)


cdominguezg:
@mnapoli hey! could you take a look on this please?

mnapoli:
Hi, I'm sorry but I don't think this is something we want to add to Lift. We don't want to add every option available to constructs.

cdominguezg:
I see, so do you have any recommendation to implement this apart of forking?

mnapoli:
I would recommend to "eject" from Lift: https://github.com/getlift/lift#ejecting

Issue #364, opened 2023-09-28
Pass origin path in REST scenario


jaulz:
### Start from the Use-case

Unfortunately, I still need to use the REST protocol of the API Gateway (because v2 is not available in Switzerland) but that means that there is an ugly stage prefix in the URL (even with a custom domain). It would be nice if we could pass the origin path to the default origin of the CloudFront distribution so we can hide the stage prefix from end users. 

### Example Config

```
service: ${env:APP_ID}

frameworkVersion: '3'

configValidationMode: error

custom:
  rootDomain: example.org
  appName: Test
  appId: ${self:service}
  domain: ${self:custom.appId}.${self:custom.rootDomain}

plugins:
    - ./vendor/bref/bref
    - serverless-lift

provider:
    name: aws
    region: eu-central-2
    stage: production
    memorySize: 1024

package:
    # Files and directories to include/exclude from deployment
    patterns:
        - '!.husky/**'
        - '!.vscode/**'
        - '!.yarn/**'
        - '!node_modules/**'
        - '!public/storage'
        - '!resources/assets/**'
        - '!storage/**'
        - '!tests/**'

functions:
    # This function runs the Laravel website/API
    web:
        handler: public/index.php
        runtime: php-81-fpm
        timeout: 28 # in seconds (API Gateway has a timeout of 29 seconds)
        events:
            - http: 'ANY /'
            - http: 'ANY /{proxy+}'

constructs:
    app:
        type: server-side-website
        domain: ${self:custom.domain}
        certificate: ...
        # Either "rest" (v1) or "http" (v2, the default) but right now only v1 is available in Switzerland
        apiGateway: 'rest'
        assets:
            '/build/*': public/build
            '/css/*': public/css
            '/fonts/*': public/fonts
            '/js/*': public/js
            '/favicon.ico': public/favicon.ico
            '/robots.txt': public/robots.txt

resources:
    Resources:
        Storage:
            Type: AWS::S3::Bucket
        CacheTable:
            Type: AWS::DynamoDB::Table
            Properties:
                AttributeDefinitions: # only keys are defined here, other attributes are dynamic
                    -   AttributeName: id # adds a mandatory id field
                        AttributeType: S # the type of id is a string
                BillingMode: PAY_PER_REQUEST # billed for each request instead of paying for a constant capacity
                TimeToLiveSpecification: # deletes cache keys automatically based on a ttl field which contains a timestamp
                    AttributeName: ttl
                    Enabled: true
                KeySchema:
[...]

PR #365, opened 2023-09-29
fix: hide stage in URL in REST scenario

jaulz:
This PR attempts to solve https://github.com/getlift/lift/issues/364. The issue is that the stage needs to be appended to the URL in REST scenario in order to view the deployed app. By using [`originPath`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-origin.html#:~:text=%3A%20No%20interruption-,OriginPath,Update%20requires%3A%20No%20interruption,-OriginShield) of the CloudFront distribution we can hide it from the user so he can simply access the CloudFront URL/custom domain.

This is my first time working on a serverless plugin so I hope I used the correct approach. Please me know if I should change anything.

### DIFF
--- a/src/constructs/aws/ServerSideWebsite.ts
+++ b/src/constructs/aws/ServerSideWebsite.ts
@@ -129,7 +129,7 @@ export class ServerSideWebsite extends AwsConstruct {
                 : undefined;
 
         // Hide the stage in the URL in REST scenario
-        const originPath = configuration.apiGateway === "rest" ? "/" + provider.getStage() : undefined;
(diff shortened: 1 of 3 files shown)


mnapoli (approved):
That LGTM, what do you think @fredericbarthelet?

jaulz:
@mnapoli @fredericbarthelet is there anything I can do to push this forward? 😊

mnapoli:
All good thanks!

PR #368, opened 2023-10-22
Use lambda url as cloudfront origin if functionName is defined

kevincerro:
This PR is a draft for using lambda URL as CloudFront origin domain

Simply remove `events` key from function and add `url: true`
Before:
`
functions:
    web:
        handler: public/index.php
        memorySize: 1769
        timeout: 28 # in seconds (API Gateway has a timeout of 29 seconds)
        runtime: php-82-fpm
        events:
            - httpApi: '*'
`

After
`
functions:
    web:
        handler: public/index.php
        memorySize: 1769
        timeout: 28 # in seconds (API Gateway has a timeout of 29 seconds)
        runtime: php-82-fpm
        url: true

constructs:
    website:
        type: server-side-website
        domain: ${param:domain}
        certificate: ${param:certificate}
        functionName: web
        assets:
            '/assets/*': public/assets
`

### DIFF
--- a/src/constructs/aws/ServerSideWebsite.ts
+++ b/src/constructs/aws/ServerSideWebsite.ts
@@ -37,6 +37,7 @@ const SCHEMA = {
     properties: {
         type: { const: "server-side-website" },
         apiGateway: { enum: ["http", "rest"] },
+        functionName: { type: "string" },
         assets: {
             type: "object",
             additionalProperties: { type: "string" },
@@ -115,11 +116,23 @@ export class ServerSideWebsite extends AwsConstruct {
         })();
         const backendCachePolicy = CachePolicy.CACHING_DISABLED;
 
-        const apiId =
-            configuration.apiGateway === "rest"
-                ? this.provider.naming.getRestApiLogicalId()
-                : this.provider.naming.getHttpApiLogicalId();
-        const apiGatewayDomain = Fn.join(".", [Fn.ref(apiId), `execute-api.${this.provider.region}.amazonaws.com`]);
+        // Resolve domain to use as cloudfront origin
+        const originDomain = (() => {
+            if (configuration.functionName !== void 0) {
+                // Use Lambda Url
+                const lambdaUrlId = this.provider.naming.getLambdaFunctionUrlLogicalId(configuration.functionName);
+
+                return Fn.select(2, Fn.split("/", Fn.getAtt(lambdaUrlId, "FunctionUrl").toString()));
+            } else {
+                // Use API Gateway
+                const apiId =
+                    configuration.apiGateway === "rest"
+                        ? this.provider.naming.getRestApiLogicalId()
+                        : this.provider.naming.getHttpApiLogicalId();
+
+                return Fn.join(".", [Fn.ref(apiId), `execute-api.${this.provider.region}.amazonaws.com`]);
+            }
+        })();
 
         // Cast the domains to an array
         this.domains = configuration.domain !== undefined ? flatten([configuration.domain]) : undefined;
@@ -132,7 +145,7 @@ export class ServerSideWebsite extends AwsConstruct {
             comment: `${provider.stackName} ${id} website CDN`,
             defaultBehavior: {
                 // Origins are where CloudFront fetches content
-                origin: new HttpOrigin(apiGatewayDomain, {
+                origin: new HttpOrigin(originDomain, {
                     // API Gateway only supports HTTPS
                     protocolPolicy: OriginProtocolPolicy.HTTPS_ONLY,
                 }),
--- a/src/providers/AwsProvider.ts
+++ b/src/providers/AwsProvider.ts
@@ -62,6 +62,7 @@ export class AwsProvider implements ProviderInterface {
     public naming: {
         getStackName: () => string;
         getLambdaLogicalId: (functionName: string) => string;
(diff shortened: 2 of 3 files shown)


src/constructs/aws/ServerSideWebsite.ts
```diff
@@ -115,11 +116,23 @@ export class ServerSideWebsite extends AwsConstruct {
...
-                ? this.provider.naming.getRestApiLogicalId()
-                : this.provider.naming.getHttpApiLogicalId();
-        const apiGatewayDomain = Fn.join(".", [Fn.ref(apiId), `execute-api.${this.provider.region}.amazonaws.com`]);
+        // Resolve domain to use as cloudfront origin
+        const originDomain = (() => {
+            if (configuration.functionName !== void 0) {
+                // Use Lambda Url
+                const lambdaUrlId = this.provider.naming.getLambdaFunctionUrlLogicalId(configuration.functionName);
+
+                return Fn.select(2, Fn.split("/", Fn.getAtt(lambdaUrlId, "FunctionUrl").toString()));
```
kevincerro, line 125: Here we must ensure that `url: true` is set on function used

src/constructs/aws/ServerSideWebsite.ts
```diff
@@ -115,11 +116,23 @@ export class ServerSideWebsite extends AwsConstruct {
...
+            } else {
+                // Use API Gateway
+                const apiId =
+                    configuration.apiGateway === "rest"
+                        ? this.provider.naming.getRestApiLogicalId()
+                        : this.provider.naming.getHttpApiLogicalId();
+
+                return Fn.join(".", [Fn.ref(apiId), `execute-api.${this.provider.region}.amazonaws.com`]);
```
kevincerro, line 133: Here we must ensure that `events` key is configured on at least one function

src/constructs/aws/ServerSideWebsite.ts
```diff
@@ -115,11 +116,23 @@ export class ServerSideWebsite extends AwsConstruct {
...
         const backendCachePolicy = CachePolicy.CACHING_DISABLED;
 
-        const apiId =
-            configuration.apiGateway === "rest"
-                ? this.provider.naming.getRestApiLogicalId()
-                : this.provider.naming.getHttpApiLogicalId();
-        const apiGatewayDomain = Fn.join(".", [Fn.ref(apiId), `execute-api.${this.provider.region}.amazonaws.com`]);
+        // Resolve domain to use as cloudfront origin
+        const originDomain = (() => {
+            if (configuration.functionName !== void 0) {
```
mnapoli, line 121: I wonder if this could be optional: if the config doesn't use API Gateway, and exactly 1 function has `url: true`, we could auto-select it?

kevincerro: What we need to do if there are multiple functions using url or api gateway.

For example:
1. function "web" uses url
2.function "api" uses apigw

Thats the reason behind the new "functionName", to be able to select which function we will get the url.

With APIGW this doesn't happen because when you have multiple functions with apigw only one "HttpApi" is created.

mnapoli: I definitely agree with the new option, it's just that if there's only 1 function with `url: true` then it could be automatic?

(simple defaults)

kevincerro: I agree with that.

If there is exactly one function defined, and it has `url: true` we must use that function lambda url by default

kevincerro: Maybe we can add a new config parameter to define wich endpoint to use if both are defined?
- `api-gateway`
- `lambda-url`

mnapoli: Oh I would add another condition: no API Gateway is defined.

But I'm not sure how easy it is to check 🤔 FYI maybe this code will help just a bit: https://github.com/serverless/serverless/blob/f22354e8dd45871915e21e5656055a9213f93ff0/lib/plugins/aws/info/display.js#L35 It's the code that displays the URL of the deployed application, and if there's only 1 it shows `endpoint: ...` instead of `endpoints: ...`

kevincerro:
Done some changes @mnapoli 
I din't managed to pass "ts-lint" checks on `utils/GetWebLambdaFunctions.ts` file after some hours, sorry 😢 .
Maybe something is wrong with AWS function events typescript definition.

src/constructs/aws/ServerSideWebsite.ts
kevincerro: Added some checks without modifying this behaviour

src/constructs/aws/ServerSideWebsite.ts
kevincerro: Added some checks

kevincerro:
@mnapoli Can you review last changes?
Thanks! 👍🏻

dnl4:
Hello @kevincerro, there is any update on this? This change will for sure reduce costs.

kevincerro:
> Hello @kevincerro, there is any update on this? This change will for sure reduce costs.

Hello, no changes since last comment.

I've a forked version of lift package with this PR changes => `npm install --require-dev @kevinc123/serverless-lift`
Working in my prod setup for some months. Will be amazin if you can try them also

PR #381, opened 2024-01-18
Set removal policy of dynamo db tables

cmcnicholas:
This feature adds support for setting the removal policy of a dynamodb table.

It is common in CI environments to setup/teardown constructs, serverless can be particularly difficult to test due to requiring AWS stack to correctly run an E2E test. Currently it is not possible to remove a serverless stack and have the database removed via lift as the default strategy is to retain the database (to avoid accidental deletion), this default is sensible.

**The default should always be used for production or persistent environments.** The optional property here allows developers to opt into automatic deletion which for CI makes life easier rather than having to script clean up of databases separately.


mnapoli (approved):
LGTM, however the tests are failing, but I'm afraid it might be because tests on master are failing :/

Issue #369, opened 2023-10-26
Control how the CloudFront cache is invalidated


Nyholm:
### Start from the Use-case

I am running a single page react app. I have 10 different JS files (chunks) after I build my app. 

It is normal that a user request 3-5 js files while they are using a part of the application. When they browse to another part they will download more files. 

There is an issue if I decide to deploy a new version. Then the additional files are removed from the S3 and CDN. Now when the user asks for the additional files they get a 404. 

I suggest to allow us to keep some files on S3 or CloudFront.
Of course I would be open for alternative solutions

mnapoli:
I wonder if we could (and if S3 allows it), instead of deleting "obsolete" files, set them to expire (be deleted) after 1 day 🤔 I haven't researched that at all but that would be a solution that we could apply by default for everyone?

jaulz:
Probably related to https://github.com/getlift/lift/issues/398

jaulz:
@mnapoli yep, I was researching on the very same right now. I wonder if we could tag all obsolete files and then use a lifecycle rule to delete these tagged files after a day. The question is whether there is an efficient way to compare and add those tags.

mnapoli:
Why not, as long as it works

Issue #398, opened 2024-06-22
Lambda is updated before assets are uploaded


jaulz:
### Description

Currently, assets are uploaded after Lambda is updated. This means that there is a short period of time where Lambda references to paths that do not exist yet and result in 404s. Is this a conscious design decision?

### How to Reproduce

```yml
service: my-app
provider:
    name: aws

functions:
    home:
        handler: home.handler
        events:
            -   httpApi: 'GET /'

constructs:
    website:
        type: server-side-website
        assets:
            '/css/*': public/css
            '/js/*': public/js

plugins:
    - serverless-lift
  ```

mnapoli:
Yes, this was an accepted compromise from the start. Building a zero-downtime deployment would be much more complex unfortunately.

jaulz:
Oh, too bad. Do you have any ideas how it could be implemented in user land instead?

jaulz:
@mnapoli I wonder if we could make the path that is about to be invalidated configurable: 
https://github.com/getlift/lift/blob/0d56286371e98fa921a88635a0b98cd3404eb488/src/classes/aws.ts#L45

In my case it's actually just the `manifest.json` which needs to be invalidated and the rest can live as long as they are cached.

mnapoli:
Ah, we're trying as hard as possible to avoid adding new options. I'm not sure if there are good alternatives for this problem 🤔

jaulz:
I'm running a server side website (Laravel) with built assets (Vite). Since assets will be uploaded afterwards, the new app version must
1. use/refer to old assets as long as the new ones are not available
2. and old assets of course must also be still available as long as the app refers to these. 

By tweaking the Vite asset resolver in Laravel and using Redis I am at least able to still refer to old assets. However, CloudFormation currently invalidates all files and hence old files are immediately not available anymore. 

I don't think that it will be possible to find a generic solution for the server side website case because the asset resolver of the app is key (which can obviously not be changed by this library). However, I think by configuring the paths that are about to be invalidated we could at least have a solution that not only works for this scenario but also for other scenarios where manifest files are used (SPA).

mnapoli:
Now that I think about it, i don't think the invalidation is the problem: there is no guarantee that older files are in cache (everywhere on the planet).

To solve this 100%, the files must exist in S3 while they are still used.

To be clear what I'm saying is that I don't think changing the CloudFront invalidation is the right solution, does that make sense?

jaulz:
[...]

PR #399, opened 2024-07-19
feat: tag obsolete files instead of deleting them

jaulz:
This change tags obsolete files instead of deleting them (see https://github.com/getlift/lift/issues/369). We use `copyObject` because it resets the time when the object was created and hence triggers the start of the lifecycle rule. Combined with a proper S3 bucket lifecycle rule we can automatically delete these tagged files after a day.


mnapoli:
I resolved conflicts and covered a missing case, but thinking about it I have a couple notes:

1. this will slow down the deployment as we need extra S3 HTTP requests -> this can amplify https://github.com/getlift/lift/issues/398 but I think this is acceptable still
2. this requires new IAM permissions on deployment (S3 tag-related permissions) -> I'm afraid this might break deployment across users 🤔

jaulz:
@mnapoli thanks for takint care of it! Would it maybe be an option to toggle it via a command flag?

mnapoli:
I'm exploring solving this plus #398 (Laravel referencing new assets before they are live) in https://github.com/getlift/lift/pull/448

PR #400, opened 2024-07-20
Chunk deleted files into batches

lepikhinb:
S3 has a batch delete limit of 1000 files.

The PR introduces chunking `keysToDelete` into batches of 1000 files to bypass the limitation.


mnapoli:
Thanks! It seems the build is failing, but I don't think it's related to the PR 🤔

mnapoli:
Have you been able to test these changes against AWS and see if they work? I'm happy to merge if this was tested manually.

mnapoli:
Got it thanks!

Issue #350, opened 2023-07-27
Add possibility to replace CloudFront functions in static-website and single-page-app constructs


InvisibleKind:
### Start from the Use-case

I want to override generated CloudFormation functions for `static-website` and `single-page-app` constructs and add custom code, like Basic Auth, custom headers and so on.

### Example Config

For that purpose I create two files: cloudfrontRequestFunctions.js and cloudfrontResponseFunctions.js and reference them under `resources` in our serverless.ts:

```
    resources: {
        Resources: {
            defaultRequestFunction: {
                Type: "AWS::CloudFront::Function",
                Properties: {
                    AutoPublish: true,
                    FunctionCode: "${file(./cf_default/cloudfrontRequestFunctions.js)}",
                    FunctionConfig: {
                        Comment: "Request handler for ${sls:stage} environment",
                        Runtime: "cloudfront-js-1.0",
                    },
                    Name: "${self:service}-${sls:stage}-request",
                },
            },
            defaultResponseFunction: {
                Type: "AWS::CloudFront::Function",
                Properties: {
                    AutoPublish: true,
                    FunctionCode: "${file(./cf_default/cloudfrontResponseFunctions.js)}",
                    FunctionConfig: {
                        Comment: "Response handler for ${sls:stage} environment",
                        Runtime: "cloudfront-js-1.0",
                    },
                    Name: "${self:service}-${sls:stage}-response",
                },
            },
        },
```

And I use the `extensions` feature of constructs to override the functions:
```
    constructs: {
        build_public: {
            type: "single-page-app",
            path: "app/dist",

            extensions: {
                distribution: {
                    Properties: {
                        DistributionConfig: {
                            DefaultCacheBehavior: {
                                FunctionAssociations: [
                                    {
                                        EventType: "viewer-request",
                                        FunctionARN: {
                                            "Fn::GetAtt": ["defaultRequestFunction", "FunctionARN"],
                                        },
                                    },
                                    {
                                        EventType: "viewer-response",
                                        FunctionARN: {
                                            "Fn::GetAtt": ["defaultResponseFunction", "FunctionARN"],
                                        },
                                    },
                                ],
                            },
                        },
                    }
                }
            }
        }
    },
```

This works just fine, but the original CloudFront Functions from Lift still exist in the CloudFormation distribution and are created as well.

[...]

PR #406, opened 2024-09-07
fix: Remove CloudFront functions

### LINKED ISSUE #129: Move from Cloudfront functions to Response headers for security
AWS announced today support for Response headers with a built-in preset for adding security headers.

https://aws.amazon.com/fr/about-aws/whats-new/2021/11/amazon-cloudfront-supports-cors-security-custom-http-response-headers/

I think it could replace the response cloudfront function and save on the extra cost. 

Unfortunatly, Cloudformation (and so CDK) support is not there yet, so I'm opening this to gather feedback and hopefully not forget about it :sweat_smile:

jackylamhk:
Fixes #129 #350

- Use 403/404 error responses to replace the request function
- Use response headers to replace the response function


jackylamhk:
Not sure how tests are run, I keep getting this error:

```
 FAIL  test/unit/permissions.test.ts
  ● Test suite failed to run

    TypeError: Converting circular structure to JSON
        --> starting at object with constructor 'Serverless'
        |     property 'yamlParser' -> object with constructor 'YamlParser'
        --- property 'serverless' closes the circle
        at stringify (<anonymous>)

      at messageParent (node_modules/jest-worker/build/workers/messageParent.js:34:19)
```

mnapoli:
Thanks for the PR!

Yeah unfortunately tests are broken on master because of changes in serverless framework, haven't found time (or even have an idea) to fix it :/

> Use 403/404 error responses to replace the request function

Could you explain this? I'm not sure what the impact is? Will the application behave any differently?

> Use response headers policy to replace the response function

Are the same exact security headers added? I.e. are there any practical changes for applications already deployed with Lift?

jackylamhk:
@mnapoli No worries.

> > Use 403/404 error responses to replace the request function
> 
> Could you explain this? I'm not sure what the impact is? Will the application behave any differently?

SPAs need their origin paths rewritten to `/index.html` to load the entry point, Lift currently uses a [CloudFront function](https://github.com/getlift/lift/blob/08772c56cc219ca7fa65cc4ff7c4dfea20246a70/src/constructs/aws/SinglePageApp.ts#L48-L60) to redirect any paths not ending in a whitelisted extension.

While this works, the function is not necessary - by setting the custom error responses for 404 (not found) and 403 (unauthorized, S3 returns one if the file doesn't exist and `s3:ListBucket` isn't allowed) we can redirect these to the entry point too.

> > Use response headers policy to replace the response function
> 
> Are the same exact security headers added? I.e. are there any practical changes for applications already deployed with Lift?

Here are the [new headers](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-response-headers-policies.html#managed-response-headers-policies-security) (also linked in docs):

| Header                      | Value                             |
| --------------------------- | --------------------------------- |
| `Referrer-Policy`           | `strict-origin-when-cross-origin` |
| `Strict-Transport-Security` | `max-age=31536000`                |
| `X-Content-Type-Options`    | `nosniff`                         |
| `X-Frame-Options`           | `SAMEORIGIN`                      |
| `X-XSS-Protection`          | `1; mode=block`                   |

And here are the [old headers](https://github.com/getlift/lift/blob/08772c56cc219ca7fa65cc4ff7c4dfea20246a70/src/constructs/aws/abstracts/StaticWebsiteAbstract.ts#L329-L334):

| Header                      | Value              |
| --------------------------- | ------------------ |
| `x-frame-options`           | `SAMEORIGIN`       |
| `x-content-type-options`    | `nosniff`          |
| `x-xss-protection`          | `1; mode=block`    |
| `strict-transport-security` | `max-age=63072000` |

So it is slightly different: mainly adding the `Referrer-Policy`.

I didn't add back the option to remove `x-frame-options` as folks should use the new `responseHeadersPolicy` key to create a custom policy with `securityHeadersBehavior.contentSecurityPolicy` instead - for security, instead of allowing all iframes. I could add it back for backward compatibility, though it might be easier to use custom policies if we keep it.

FYI - this build was tested with a SPA and Serverless deployment.

PR #416, opened 2025-01-20
Fixing tests

jiw-mh:
I noticed that the tests were failing, I believe this PR should fix the tests.

PR #417, opened 2025-01-20
Feature: Option workerRef

jiw-mh:
This is a stab at a reference defining the worker reference following the discussion in https://github.com/getlift/lift/discussions/415


mnapoli:
Thanks!

I wonder if this could work instead:

```yml
functions:
    foo:
        handler: worker.handler

constructs:
    emails:
        type: queue
        worker: foo
```

that would be simpler.

jiw-mh:
Does Serverless support varied types?

mnapoli:
This pattern is very common in serverless.yml (string or object value).

jiw-mh:
I think I figured out how to specify the function and updated the PR to make worker optionally as a string

test/fixtures/queuesWorkerRef/serverless.yml
```diff
@@ -0,0 +1,16 @@
...
+    versionFunctions: false
+
+functions:
+    foo:
+        handler: worker.handler
+
+constructs:
+    emails:
+        type: queue
+        workerRef: foo
```
mnapoli, line 16:
```suggestion
        worker: foo
```

?

mnapoli:
Can you document this new feature here: https://github.com/getlift/lift/blob/master/docs/queue.md#worker

Also I think it might be worth rebasing after we merge #416 so that we get a clear picture of the CI.

jiw-mh:
I added some documentation and rebased the branch.

jiw-mh:
@mnapoli is there something else I can do?

mnapoli:
Thanks!

PR #425, opened 2025-10-06
move dep to open source serverless

dan-greene-brivo:
Moving package.json dependency to [osls](https://github.com/oss-serverless/serverless) from serverless as serverless v3 is not really supported anymore and osls will receive security vulnerability updates.  Also added @types/lodash as it needed it to build (don't know how/why it wasn't already there)


mnapoli:
This only changes the dev dependencies and not the `peerDependencies`. I'm not sure if we should change both or if it could cause issues with some people 🤔

dan-greene-brivo:
> This only changes the dev dependencies and not the `peerDependencies`. I'm not sure if we should change both or if it could cause issues with some people 🤔

Good catch - I'll update the peerDependency as well - unfortunately- I think there's going to be a 'line in the sand' that will be needed to take on official serverless v4 or open source v3.  What do you think of putting both in peerDependency and flagging them both as optional via peerDependenciesMeta ?

dan-greene-brivo:
to be honest - it was so long ago - I don't remember - I'll try to trigger another test run

mnapoli:
Done in #445 thank you!

PR #426, opened 2025-11-05
feat: add `.lottie` to the redirect rgx for `single-page-app` construct

nolde:
[dotLottie files](https://en.wikipedia.org/wiki/Lottie_(file_format)#The_dotLottie_file_format) are a growing option for online animations.

I have a website full of those files. I am adding `.lottie` file to the redirect regex for `single-page-app` constructs. This allows lottie files to work properly when deployed by lift.


nolde:
Indeed, there are two tests affected. My bad.

The two tests affected by the change have been updated.

However, there are many tests failing that are unrelated to my changes, so I believe any new test runs will still fail.

